PR #66, opened 2026-06-29
fix(rails): inject class_methods block self-type into the analyzer's own type-check

felixefelip:
## Problem

A method inside an `ActiveSupport::Concern`'s `class_methods do … end` block that calls an implicit-self scope or class method — e.g. a scope defined in `included do`, or AR query methods like `joins` — was inferred as `untyped`:

```ruby
module Card::Entropic
  extend ActiveSupport::Concern
  included do
    scope :due_to_be_postponed, -> { … }
  end
  class_methods do
    def auto_postpone_all_due
      due_to_be_postponed.find_each { |c| … }   # => inferred `() -> untyped`
    end
  end
end
```

`steep check`/langserver resolve `find_each` to `-> nil` here, but `rbs_infer` emitted `untyped`.

## Root cause

`ClassMethodsExpander` desugars `class_methods do` into a nested `module ClassMethods` for RBS generation. Those methods run with `self` = the includer's singleton (`singleton(Card)`), so `due_to_be_postponed` (a singleton method on `Card`) only resolves when the submodule carries that self-type.

The downstream `.steep_module_self_types.yml` sidecar already supplies it through its `blocks` `self` entry (`singleton(::Card) & ::Card::Entropic::ClassMethods`) — that's why `steep check`/langserver resolve the chain. But the analyzer's **own** type-check (`SteepBridge`, used to infer return types during generation) never received it: `Analyzer#load_and_parse_target` injected only the *concern-level* self-types, omitting the `class_methods` block self. So inside the desugared `module ClassMethods`, `self` was just the bare module, `due_to_be_postponed` resolved to `untyped`, and the whole `.find_each` chain — and the method return — collapsed to `untyped`.

## Fix

Inject the same `singleton(<includer>) & ::<FQN>::ClassMethods` self onto the desugared `module ClassMethods` during `load_and_parse_target`, via `Steep::Source::ModuleSelfTypes.inject`. The block is detected from the **pre-expansion** source, since expansion already removed the `class_methods` call the detector keys on.

The self-type computation is reused from the existing `ClassMethodsImplements` (it already produced this exact string for the sidecar) through a new `self_type_entry` helper — the core learns no Rails DSL; the framework knowledge stays in `extensions/rails/`.

## Verification

The existing dummy fixture `Post::Taggable` had the same bug — `tags_ordered_by_tag_name` calls `joins(:tag).order(...)` (a singleton method on `Post`):

```diff
-      def tags_ordered_by_tag_name: () -> untyped
+      def tags_ordered_by_tag_name: () -> Post::ActiveRecord_Relation
```

- `spec/integration/` → 53 examples, 0 failures (snapshot regenerated).
- `class_methods_implements_spec` (3 new cases for `self_type_entry`) + `analyzer_spec` → green.

### DIFF
--- a/docs/engineering/keep-core-framework-agnostic.md
+++ b/docs/engineering/keep-core-framework-agnostic.md
@@ -15,11 +15,19 @@ existing plugin seams so the core stays oblivious:
   DSL is expressible as plain Ruby the core already handles. Examples:
   `CurrentAttributesExpander`, `OnLoadExpander`, `ClassMethodsExpander`
   (`class_methods do … end` → `module ClassMethods … end`).
+- **`RbsInfer::Project::SelfTypeAnnotators`** — inject `@type self:` / `@type instance:`
+  comments into the source *after* expansion but *before* the parse, for cases that aren't a
+  plain-Ruby desugaring but a self-type the pipeline (and Steep, as the return-type oracle)
+  must see — e.g. a concern's host-class self, or the includer-singleton self of a desugared
+  `module ClassMethods`. An annotator implements
+  `self_type_entries(path:, module_name:, source:) -> [{ "anchor", "annotations" }]`; the core
+  places each via Steep's generic `Source::ModuleSelfTypes.inject`, never naming the extension.
+  Examples: `ModuleSelfTypeAnnotator`, `ClassMethodsImplements`.
 - **Generators** (`enumerize/`, `rails/custom_generator.rb`, `erb_convention_generator.rb`,
   `devise/`, `carrierwave/`) — emit standalone sidecar RBS into a dedicated `sig/` dir via a
   rake task, never touching the core analysis of the source file.
 
-Both seams let an extension register itself at require time; the core knows none of them.
+All three seams let an extension register itself at require time; the core knows none of them.
 
 Litmus test: *would this `when`/`if`/constant-name make sense for a gem that doesn't exist?*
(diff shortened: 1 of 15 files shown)


felixefelip:
## Follow-up: second root cause (the `nil` filter)

After the self-type injection landed, `make rbs_infer_all` on a real app fixed several `class_methods` returns but **not** the reported `auto_postpone_all_due` (`scope.find_each { … }`). Investigation showed the self-type injection was working — the scope *did* resolve and Steep correctly typed the body as `nil` — but the return-type resolver was **discarding every `nil` Steep inferred** (`steep_type != "nil"` guard), leaving the method `untyped`.

Reproduced in the dummy with a `recently_tagged` scope in `included do` called from a `class_methods` iterator:

```diff
-      def touch_recently_tagged: () -> untyped
+      def touch_recently_tagged: () -> nil
```

The env is built with `implicitly_returns_nil: false`, so a `nil` from Steep is genuine, not a fallback — so it's now emitted. The only unsafe case is a **conditional tail** (`X if cond`, `case` without a value `else`): when the value branch is `untyped`, Steep collapses `untyped | nil` → `nil`, and `-> nil` would hide it (e.g. `posts.destroy_all if cond`). `unconditional_nil_tail?` guards that, keeping such methods `untyped`.

Verified: dummy `steep check` baseline still passes on the generated `-> nil`; `untag` (conditional tail) correctly stays `untyped`.

PR #73, opened 2026-07-03
feat: desugar belongs_to default: + association construction into example.rb-shaped Ruby (#72)

### LINKED ISSUE #72: Rails extension: desugar belongs_to default: + association construction into example.rb-shaped Ruby (nilable-by-flow)
Generation side of the Fizzy `belongs_to :x, default: -> { y.z }` false-positive fix. Steep's consuming side (type-check the expansion + map diagnostics back to the `default:` lambda) is the companion: felixefelip/steep#54.

## The problem

```ruby
class Assignment < ApplicationRecord
  belongs_to :post
  belongs_to :owner, class_name: "User", default: -> { post.user }
end
```

`post` is a required `belongs_to` typed `Post?`, so `steep check` flags the default lambda:

```
app/models/assignment.rb:19: Type `(::Post | nil)` does not have method `user`
```

Runtime never raises: an `Assignment` is only ever built through `post.assignments` (which sets `post`), and `default:` runs in `before_validation`. See the checked-in fixture `spec/dummy/app/models/assignment.rb` (model) + `app/controllers/posts/assignments_controller.rb` (the `@post.assignments.create!` safe path).

## Key insight (why this is NOT a CurrentAttributes-style expander)

The CurrentAttributes `SourceExpander` desugars for **rbs_infer's own inference only** — its comment says it "is never … seen by the app's `steep check`". That works there because the fix is purely about **inferring a type**. Here the RBS is already correct (`post: Post?`); the problem is Steep's **check of the real lambda**. So this expansion must be **consumed by Steep** (via #54), not just fed to rbs_infer's parse.

## The approach: emit example.rb-shaped Ruby (flow-based, no `Validated`)

We already proved end-to-end (see `spec/integration/steep_scenarios_spec.rb`) that the contract mechanism (#51 explicit-receiver + attribute-write narrowing, #52 transitive closure) resolves the `example.rb` shape:

```ruby
column = Column.new(...); column.board = board; column.save   # board non-nil -> save chain enforces
```

So the extension emits the AR flow **as that exact shape**, plus a **source-map**, for Steep (#54) to check and map back.

### 1. Model side — `belongs_to default:` → lifecycle methods

```ruby
def save; run_before_validation_callbacks; end
def run_before_validation_callbacks; run_belongs_to_default_callbacks; end
def run_belongs_to_default_callbacks
  self.owner = post.user    # inline the lambda BODY (not `-> {}.call`) so the
                            # contract inferrer sees the `self.post` deref directly
end
```

- Inline the lambda body directly in a method body (the inferrer walks `def` bodies for `NoMethod` on `self.x`).
- Carry a source-map span: the inlined `self.owner = post.user` ↔ the real `default: -> { post.user }`.

### 2. Caller side — `owner.has_many_assoc.new/build/create[!]` → build + owner-setter + save

```ruby
# @post.assignments.create!(params)  ==>
a = Assignment.new(params)
a.post = @post      # the belongs_to that the default derefs, set from the association's OWNER
a.save!
```

The type of `post` flows through the **owner** of the association (not the FK, not `new`'s args):

```
Post#assignments  ->  self : Post
                  ->  <collection>.owner : Post          # proxy specialized by owner type
                  ->  a.post = owner   ->  a.post : Post  # non-nil (attribute-write narrowing, #51)
```

So the emitted collection proxy must be **specialized by the owner type** (à la the fixture's `Post_Owner_Assignment::…CollectionProxy`), so `owner` is typed `Post` and the belongs_to it sets is typed non-nil.

### Argument-typing rules (the "typed like example.rb" requirement)

The desugared `new`/`create` must carry real types so flow narrowing knows nilability per attr:
- **association attrs** → typed from the owner (`a.post = owner`, non-nil);
- **literal kwargs** at the call site (`create!(owner: some_user)`, `some_user: User`) → typed from the literal (narrows `a.owner`);
- **ParamsBag args** (`create!(assignment_params)`) → untyped → those attrs stay nilable (correct: if `owner` came from params it's nilable → the default runs → derefs `post`, which is typed non-nil via the owner).

## Where it lives

- A **Rails extension** under `lib/rbs_infer/extensions/rails/` (knows `belongs_to default:`, `has_many`, the inverse/element type — via the reflection, like the other Rails generators). Core stays framework-agnostic.
- Emits the **expanded Ruby + source-map** as a sidecar Steep consumes (#54), analogous to how the module-self-types generator emits `.steep_module_self_types.yml`.
- Reuses the existing seam pattern; the contract narrowing/enforcement is unchanged (#51/#52).

## Soundness

Because it goes through the contract mechanism (not a blind `Validated` narrow), an `Assignment.new.save` that does **not** establish `post` still errors — the false positive is removed only where construction actually proves `post` is set.

## Deliverables

- [ ] Rails extension emitting the model-lifecycle expansion (belongs_to `default:` → inlined callback methods) + source-map.
- [ ] Association-construction expansion (`owner.assoc.new/build/create[!]` → build + owner-setter + save), with the owner-typed collection proxy.
- [ ] Argument-typing rules (association / literal kwargs / ParamsBag).
- [ ] End-to-end coverage in the dummy (the `Assignment` + `Posts::AssignmentsController` fixtures) — the `(::Post | nil)` baseline error should clear once #54 lands.

felixefelip:
Closes #72 (generation side). Companion: felixefelip/steep#54 (the consuming side — type-check the expansion + map diagnostics back to the `default:` lambda).

## The problem

```ruby
class Assignment < ApplicationRecord
  belongs_to :post
  belongs_to :owner, class_name: "User", default: -> { post.user }
end
```

`post` is a required `belongs_to` typed `Post?`, so `steep check` flags the default lambda:

```
app/models/assignment.rb:19: Type `(::Post | nil)` does not have method `user`
```

Runtime never raises — an `Assignment` is only ever built through `post.assignments` (which sets `post`), and `default:` runs in `before_validation`. The generated RBS is already correct; the false positive is Steep's check of the **real lambda**. So this must be **consumed by Steep** (#54), not fed to rbs_infer's parse — which is why it's a generator emitting a sidecar, not a `SourceExpander`.

## The approach

`RbsInfer::Extensions::Rails::BelongsToDefaultGenerator` emits the AR flow as the proven `example.rb` contract shape (steep #51 explicit-receiver + attribute-write narrowing, #52 transitive closure):

- **Model side** — inlines the lambda body into a `save`-reachable callback:
  ```ruby
  def save; run_before_validation_callbacks; end
  def run_before_validation_callbacks; run_belongs_to_default_callbacks; end
  def run_belongs_to_default_callbacks
    self.owner = post.user      # lambda BODY inlined, source-mapped to the real default:
  end
  ```
- **Caller side** — `@post.assignments.create!(...)` → build + owner-setter + save, the type flowing through the association **owner**:
  ```ruby
  record = Assignment.new
  owner  = Post.new
  record.post = owner           # inverse belongs_to, non-nil (#51)
  record.save                   # precondition satisfied → no error
  ```
- **Argument-typing rules** — association attrs (from owner), non-nil literal kwargs (narrow), ParamsBag args (stay nilable → deref still flagged, soundly).

Two sidecars mirror the `.steep_module_self_types.yml` precedent: `.steep_belongs_to_default.rb` (the program Steep checks) and `.steep_belongs_to_default.yml` (source-map back to the real `default:` lambda).

## Verified end-to-end

Ran the emitted expansion through the actual rbs_infer → Steep contract pipeline:

- **Safe path** (`@post.assignments.create!`): `post` inferred `...Post?`, preconditions inferred, **0 diagnostics** → native `assignment.rb:19` diagnostic gets suppressed once #54 lands.
- **Unsafe path** (`Assignment.create!(params)` reaching `save` without `post`): still errors `(::...Post | nil) does not have method user` → mapped back to the real lambda. **Soundness holds.**

Tests: 10 unit specs + 2 end-to-end Steep scenarios; full default suite **700 examples, 0 failures**.

## Deliverables

- [x] Rails extension emitting the model-lifecycle expansion + source-map
- [x] Association-construction expansion (owner-typed setter path)
- [x] Argument-typing rules (association / literal kwargs / ParamsBag)
- [x] End-to-end coverage — the `(::Post | nil)` baseline error clears on the safe path (pending #54 to suppress the native diagnostic)

## Notes

- The `steep_baseline.txt` line stays until #54 lands — this PR is the generation half; it doesn't change native Steep output.
- Discovered a latent analyzer bug: an underscore in a constant name defeats `.new`/external-setter resolution (types attrs `untyped`). Worked around with camelCase synthetic names; worth a dedicated fix later.


lib/rbs_infer/extensions/rails/belongs_to_default_generator.rb
```diff
@@ -0,0 +1,133 @@
...
+      #     lambda, the line in the expansion that stands in for it, so Steep
+      #     (#54) remaps any expansion diagnostic to the real span AND
+      #     suppresses its native check of that span (now covered here).
+      class BelongsToDefaultGenerator
+        SIDECAR_PATH = "sig/generated/.steep_belongs_to_default.yml"
+        EXPANDED_PATH = "sig/generated/.steep_belongs_to_default.rb"
+
+        # Roots scanned for models (reflections) and for construction sites.
+        MODEL_ROOTS = %w[app/models].freeze
+        CALLER_ROOTS = %w[app/controllers app/models app/services].freeze
```
felixefelip: It looks bad

lib/rbs_infer/extensions/rails/belongs_to_default_generator.rb
```diff
@@ -0,0 +1,133 @@
...
+        end
+
+        private
+
+        def scan_models
+          each_rb(MODEL_ROOTS).filter_map do |abs, rel|
+            BelongsToDefault::ReflectionScanner.scan(path: rel, source: File.read(abs))
+          rescue StandardError
+            nil
+          end.flatten
```
felixefelip: it's bad to debug

lib/rbs_infer/extensions/rails/belongs_to_default/construction_site_scanner.rb
```diff
@@ -0,0 +1,173 @@
+# frozen_string_literal: true
+
+require "prism"
+
+module RbsInfer
+  module Extensions
+    module Rails
```
felixefelip: Add the module ActiveRecord

spec/dummy/sig/generated/.steep_belongs_to_default.rb
```diff
@@ -0,0 +1,36 @@
...
+class RbsInferBelongsToDefaultPost
+  def user; raise; end
+end
+
+class RbsInferBelongsToDefaultUser
+end
+
+class RbsInferBelongsToDefaultAssignment
+  attr_accessor :post, :owner
+  def save; run_before_validation_callbacks; end
```
felixefelip: I think is better to read the methods with break lines instead of inline

spec/dummy/sig/generated/.steep_belongs_to_default.rb
```diff
@@ -0,0 +1,36 @@
...
+# Regenerated on every run; do not edit.
+
+class RbsInferBelongsToDefaultPost
+  def user; raise; end
+end
+
+class RbsInferBelongsToDefaultUser
+end
+
+class RbsInferBelongsToDefaultAssignment
```
felixefelip: I think it would be better to have a file for each class

felixefelip:
Superado pela nova direção (**Forma 2**) acordada na #72: pseudocódigo com os **nomes reais** (`Post_Assignment::ActiveRecord_Associations_CollectionProxy`), checando o **código real** do app, com o **fluxo de `save` modelado** e os contratos (precondition/postcondition) **inferidos pelo fork do Steep** — em vez do mundo isolado (Forma 1) deste PR.

O código da Forma 1 fica **preservado no histórico** desta branch. A fonte da verdade do design passou a ser a **issue #72** (§8.1 e §8.2 decididas, `foo.rb` como contrato). Um novo PR será aberto quando a implementação da Forma 2 estiver pronta.

PR #142, opened 2026-07-30
fix: type a local argument by its narrowed type at the call site

felixefelip:
## Problem

`IntraClassCallAnalyzer#resolve_value_type` answered a `LocalVariableReadNode` from `@local_var_types` — a map of *variable name to type*, built once per method from Steep's `:lvasgn` nodes. One type per variable cannot express narrowing, which is positional. So every read got the type of the **assignment**:

```ruby
if session = find_session_by_cookie
  set_current_session session
end
```

Measured on that exact source (Steep's own typing):

```
lvasgn  line 3 col 7   -> ((::Session & ::Session::Validated) | nil)   <- what was being read
lvar    line 4 col 26  -> (::Session & ::Session::Validated)           <- what Steep already knew
```

and `set_current_session` came out as `((Session & Session::Validated)? session)`. The argument cannot be nil there — the `if` is precisely what guarantees it.

## Why it mattered downstream

A nilable parameter makes every fact about it unprovable, and the damage is not local:

- `Current.session = session` establishes nothing, because the written value may be nil. That refusal is *correct* — the fault is upstream, in the parameter's type. Measured: same body, non-nilable parameter, and the establishment appears.
- `session.signed_id`, two lines below, became a call on a nilable receiver — which is why the cookie hash kept typing as `{ value: untyped, … }` after `find_signed` was given a signature.

## Fix

Steep had the answer and simply was not asked. `SteepBridge#local_var_read_types` collects the `:lvar` nodes with the narrowing in effect at each one, and `resolve_value_type` consults the read's position before falling back to the per-method map. Positional arguments, keyword arguments and `Klass.new(x:)` all now see the type that holds where the value is actually passed.

**Keyed by line and CHARACTER column.** Prism counts columns in bytes (`start_column`) and in characters (`start_character_column`); Parser counts characters. Matching on the byte column would drift on any source with multibyte text ahead of the argument on its line. A spec pins the key.

## Left alone

`Inference::NewCallCollector#resolve_value_type` has the same shape, but it receives its locals map from the caller and deliberately **mutates** it to model flow — it stores ivars in there too. Giving it positional lookup is a change with its own semantics, not a copy of this one, so it is out of scope here.

## Verification

- `steep_bridge_spec`: the per-read map gives the narrowed type where the per-method map reports nilable, plus a test pinning the `[line, column]` key.
- `intra_class_call_analyzer_spec`: end-to-end, the callee's parameter is inferred as `(User & User::Validated)` rather than `…?`. Reverting just the lookup fails it with `got: "(User & User::Validated)?"`.
- Full suite: **817 examples, 0 failures**.

### DIFF
--- a/lib/rbs_infer/inference/caller_file_analyzer.rb
+++ b/lib/rbs_infer/inference/caller_file_analyzer.rb
@@ -56,9 +56,17 @@ module RbsInfer::Inference
 
       # Use Steep to resolve local var types (including block params)
       caller_constant_types = {}
+      # Per-READ types (felixefelip/rbs_infer#142). `local_var_types` below is
+      # flattened across every method in the file, first-wins, so one variable
+      # name carries one type for the whole FILE — coarser even than the
+      # per-method map it comes from, and blind to narrowing either way.
+      local_var_read_types = {}
+      local_var_types_by_method = {}
       if @steep_bridge
         steep_vars = @steep_bridge.local_var_types_per_method(source)
+        local_var_types_by_method = steep_vars
         steep_vars.each_value { |vars| local_var_types.merge!(vars) { |_k, old, _new| old } }
+        local_var_read_types = @steep_bridge.local_var_read_types(source)
         caller_constant_types = @steep_bridge.constant_types(source)
       end
       constant_arg_resolver = ConstantArgTypeResolver.new(
@@ -124,6 +132,8 @@ module RbsInfer::Inference
         target_class: @target_class,
         method_return_types: method_return_types,
         local_var_types: local_var_types,
+        local_var_read_types: local_var_read_types,
+        local_var_types_by_method: local_var_types_by_method,
         method_type_resolver: @method_type_resolver,
         caller_class_name: caller_class_name,
         init_positional_params: @init_positional_params,
--- a/lib/rbs_infer/inference/intra_class_call_analyzer.rb
+++ b/lib/rbs_infer/inference/intra_class_call_analyzer.rb
@@ -17,6 +17,9 @@ module RbsInfer::Inference
       @current_param_names = Set.new
       @method_positional_params = method_positional_params
       @steep_local_var_types = steep_bridge && source_code ? steep_bridge.local_var_types_per_method(source_code) : {}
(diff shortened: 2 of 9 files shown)


felixefelip:
Extended after running the generator on the app that motivated this. The first commit fixed the intra-class path and left the job half done:

```
def set_current_session: (((Session & Session::Validated) | (Session & Session::Validated)?) session)
```

That union is not two callers disagreeing. It is **one call site resolved twice**, by two paths, one of them stale: `Analyzer` unions the intra-class parameter types with the cross-class ones (`analyzer.rb:242`), and the cross-class path — `CallerFileAnalyzer` → `NewCallCollector` — still read the assignment's type. Measured by disabling the cross-class path:

```
with it:     (((Session & Session::Validated) | (Session & Session::Validated)?) session)
without it:  ((Session & Session::Validated) session)
```

Teaching `NewCallCollector` the per-read map was not enough on its own, and instrumenting it said why:

```
session line 50 -> (Session & Session::Validated)    # read hit
session line 98 -> (Session & Session::Validated)?   # read miss -> fallback
```

Line 98 passes a block parameter Steep leaves untyped, so it falls back to `local_var_types` — which `CallerFileAnalyzer` flattened across the whole file, first-wins. One type per variable NAME for an entire file, applied across method boundaries: a block parameter in one method got the type of a *different* method's local that happened to share its name. Entering a `def` now swaps the file-wide names for that method's own; a source with no `def` (an ERB template is one method's body) keeps the flat map, which is all it ever had.

So this PR now reverses the "left alone" note in its description: `NewCallCollector` is in scope, because leaving it out left the wrong type winning half the time.

**Two integration expectations move**, both toward the qualified name Steep answers with:

```diff
-    def user=: (User value) -> String?
+    def user=: (Example2::User value) -> String
-      attr_accessor user: (User | nil)?
+      attr_accessor user: (Example3::User | nil)?
```

The return going from `String?` to `String` follows: with the parameter qualified, `value.name` resolves to `Example2::User`'s `attr_reader name: String` instead of falling back to `Foo`'s nilable `name`. Both files already emitted the qualified form on neighbouring lines — the old values were the outliers, not the new ones.

Suite: **817 examples, 0 failures**. Two new unit specs on the collector, both failing with the change reverted.

PR #143, opened 2026-07-30
Refine record returns from indexed assignments

felixefelip:
## Summary
- resolve typed values inside record RHSs of indexed assignments
- preserve Ruby assignment return semantics when the writer is untyped
- add regression coverage for session.signed_id-style values

## Verification
- bundle exec rspec spec/lib/rbs_infer/inference/type_merger_spec.rb spec/lib/rbs_infer/ast/node_type_inferrer_spec.rb

### DIFF
--- a/lib/rbs_infer/inference/type_merger.rb
+++ b/lib/rbs_infer/inference/type_merger.rb
@@ -308,8 +308,20 @@ module RbsInfer::Inference
       signature.match?(/-> \{.*\buntyped\b.*\}\z/)
     end
 
+    # The trailing return type, and only that. `-> .+\z` is greedy AND
+    # leftmost, so it matched from the FIRST arrow: a signature carrying a block
+    # (`() { (Integer) -> void } -> untyped`) came out as
+    # `() { (Integer) -> String`, unbalanced and unparseable.
+    #
+    # The alternation names exactly the two shapes that can reach here — the
+    # original `-> untyped`, and the partially resolved record
+    # `-> { value: untyped }` — and both are anchored at the end, so an arrow
+    # inside a block type can never be the match. A signature matching neither
+    # is left alone rather than mangled.
+    RETURN_TYPE_SUFFIX = /-> (?:untyped|\{.*\})\z/
+
     def replace_return_type(member, type)
-      member.signature = member.signature.sub(/-> .+\z/, "-> #{RbsInfer::Signatures::RbsParserUtil.parenthesize_union(type)}")
+      member.signature = member.signature.sub(RETURN_TYPE_SUFFIX, "-> #{RbsInfer::Signatures::RbsParserUtil.parenthesize_union(type)}")
     end
 
     # Extrai nome do método quando o receiver é self implícito ou explícito
--- a/lib/rbs_infer/inference/type_merger.rb
+++ b/lib/rbs_infer/inference/type_merger.rb
@@ -131,7 +131,12 @@ module RbsInfer::Inference
         owner = collector.owner_of(defn)
         member = members.find { |m| m.kind == kind && m.name == method_name && m.owner == owner }
         next unless member
-        next unless member.signature.end_with?("-> untyped")
+        # A record assembled by the syntax-only pass may retain an `untyped`
+        # field even though the RHS of its enclosing assignment is resolvable.
+        # Keep those members in this pass so an indexed write such as
+        # `cookies[:token] = { value: session.signed_id }` can refine the
+        # record from the typed RHS rather than from the (untyped) writer.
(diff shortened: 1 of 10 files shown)


felixefelip (commented):
Reviewed, and pushed fixes for the one blocking defect I found. Everything below was measured rather than read off the diff.

## The change is right, and it delivers

The reasoning in the comment is sound: Ruby evaluates `a[k] = v` to `v` regardless of what `[]=` returns, so typing the assignment expression from its RHS is more correct than following an untyped writer. Verified on the application that motivated it:

```
def set_current_session: (...) -> { value: String, httponly: bool, same_site: Symbol }   # was value: untyped
def resume_session:      () -> { httponly: bool, same_site: Symbol, value: String }?
def require_authentication: () -> { httponly: bool, same_site: Symbol, value: String }
```

## Fixed here: `replace_return_type` corrupted block-bearing signatures

`sub(/-> .+\z/, ...)` is greedy **and** leftmost, so it matched from the *first* arrow:

```
before: def write: (Session session) { (Integer) -> void } -> untyped
after:  def write: (Session session) { (Integer) -> String
```

Unbalanced, unparseable. The `sub(/-> untyped\z/, ...)` it replaced was safe precisely because it was anchored at the end — the anchoring is what had to be kept while widening the match to cover the record shape.

The risk was not limited to records: `replace_return_type` also took over the `infer_call_return_type` branch, which fires for ordinary methods, so any block-bearing signature ending in `-> untyped` was exposed.

The existing spec meant to guard this (`"não corrompe assinatura de método com bloco..."`, renamed to `"does not corrupt a block-bearing signature..."` by the commits below) did not catch it — it only inspects `ClassMemberCollector` output and never reaches `resolve_method_return_types_from_attrs`, where the substitution happens.

Pushed as `095c4a1`: the match is now `/-> (?:untyped|\{.*\})\z/`, naming exactly the two shapes that can reach the method, both anchored, so an arrow inside a block type can never be the match. Checked against a block returning a record (`() { (Integer) -> { a: Integer } } -> untyped`), a block plus a record return, and an already-resolved signature (now left alone instead of mangled). The spec added with it drives the real path and fails with the greedy regex restored.

Two follow-up commits put the spec file's comments and example descriptions in English, including the pre-existing ones; the comment that cites another example by name was updated in step.

## Still open

**1. The full suite fails: 820 examples, 1 failure.** `spec/integration/steep_dummy_spec.rb` reports two new steep errors, from the fixtures this PR adds:

```
app/models/example14.rb:15:17: Type `(::Example14::User | nil)` does not have method `name`
app/models/example15.rb:25:25: Type `(::Example15::User | nil)` does not have method `name`
```

The PR's verification section lists two spec files, which is why this passed unnoticed. Judging by the fixtures' own comments (`# when @halted is false @user is not nil`) the errors look deliberate — reproductions of a guard that does not narrow an ivar. If so, `UPDATE_STEEP_BASELINE=1` records them; if not, the fixtures belong in their own PR, since they are unrelated to record typing.

**2. Six other branches in `type_merger.rb` still use `sub(/-> untyped\z/, ...)`** while only one uses `replace_return_type`. `unresolved_record_return?` now admits members whose signature ends in `-> { ... }`, and for those the other six substitute nothing — the record refinement only happens through the single branch that was converted. Not a bug today; it will surprise whoever expects a record refined via another path.

## Smaller observations

1. **`example14` and `example15` are not referenced by any spec** (`grep` outside `spec/dummy` finds nothing). Every other `exampleN` has an integration example pinning its output; these two will drift silently.
2. **`Example14::User#initialize(name:)` has `@name` as its body** — a read, no assignment. Looks like a typo for `@name = name`, and it is what produces `attr_reader name: untyped` in the generated sig.
3. **Inconsistent qualification inside one generated file**: `def set_user: () -> User` next to `def verify: () -> Example14::User`.
4. **`resolved_record_type` is all-or-nothing** — one unresolvable field discards the whole record. Defensible, but `{ value: String, x: untyped }` would still beat `untyped`.
5. **`unresolved_record_return?`** matches `\buntyped\b` anywhere inside a trailing record, including one nested in a block type. Harmless (resolution then fails and nothing is written), but it could be tightened.
6. **The `.steep_postconditions.yml` commit** carries churn from the *steep* side — the `Current.*` facts lifted into `__rbs_infer__run_*` by the const-establishment closure — which has nothing to do with record typing. Worth a line in the description, or a reviewer will wonder why `Current.author_name` appears in this PR.

## Caveat on my measurements

The suite run used a working tree with uncommitted local edits to `example15` (`return if @halted` and `|| halt` commented out). `example14`'s baseline error is independent of that; `example15`'s I could not attribute with certainty to the pushed state.

PR #248, opened 2026-08-15
Read a stored-block replay written in the concern's own direction

### LINKED ISSUE #247: StoredBlockReplayExpander cannot read the concern's own direction (`base.class_eval(&@block)`)
`spec/dummy/app/models/example32.rb` is the closest fixture we have to a real `ActiveSupport::Concern` — and it is the one shape `StoredBlockReplayExpander` cannot read. `Example32::Bar` gets `extend ::Example32::Foo` and nothing else; the `def age` written inside `bazingado do … end` never lands on it.

```ruby
class Example32
  module Foo
    def bazinga(module_included)
      module_included.bazingado(self)
    end

    def bazingado(base = nil, &block)
      if base.nil?
        @_bazingado_block = block
      else
        base.class_eval(&@_bazingado_block) if @_bazingado_block
      end
    end
  end

  module Baz
    extend Example32::Foo
    bazingado do
      def age = 31
    end
  end

  class Bar
    extend Example32::Foo
    bazinga(Example32::Baz)   # `age` should land here
  end
end
```

This mirrors `Concern#included(base = nil, &block)` storing into `@_included_block` and `append_features(base)` replaying it with `base.class_eval(&@_included_block)`.

## It is not the ivar

Worth stating up front, because it is the obvious suspect and it is wrong. `stored_block_ivar` (`collector.rb:116`) picks up `@_bazingado_block = block` exactly as it picks up example29's `@bazingado_block = block` — `@storages` is populated identically for both. example29/31 store in an ivar too; their `attr_reader` is just a public window onto the same slot.

Probing the collector directly:

```
== example29 (works)
  storages:      [["Example29::Foo", "bazingado", "@bazingado_block"]]
  replay_methods:[["Example29::Foo", "bazinga", "module_included", "bazingado_block"]]
  readers:       [["Example29::Foo", "bazingado_block", "@bazingado_block"]]
  => REPLAYS:    [["Example29::Bar", "bazingado"], ["Example29::BarOther", "bazingado"]]

== example32 (declines)
  storages:      [["Example32::Foo", "bazingado", "@_bazingado_block"]]
  replay_methods:[]
  readers:       []
  => REPLAYS:    []
```

### 1. `replay_shape` insists the block arrive *through the parameter*

`collector.rb:130-142` matches only `class_eval(&<local_param>.<reader>)`. A variant that keeps the ivar *and* adds an `attr_reader`, but reads it off `self` (`base.class_eval(&_blk)`), still yields `replay_methods: []`. So the axis is not "ivar vs reader" — it is *whose* state holds the block.

That is the direction inversion:

| | `self` inside the replay | the source module | the target |
|---|---|---|---|
| example29 / example31 | the **target** | the parameter (`module_included.bazingado_block`) | `self` |
| example32, real `Concern` | the **source** | `self` (`@_bazingado_block`) | the parameter (`base`) |

Once you are inside the object that holds the block, reading the ivar directly is the natural spelling — the ivar-direct read is a *consequence* of the inversion, not a separate cause. Which also means the `attr_reader` requirement (`collector.rb:193`, `reader_ivars`) has no meaning in this direction: there is nothing to expose a reader *to*.

### 2. The `class_eval` must sit in the method the target's body calls

`resolve_replays` (`collector.rb:186`) looks up `replay_by_owner_and_method[[owner, apply.method]]`, and `apply.method` is `bazinga` — the call written in `Bar`. example32 splits the work: `bazinga` forwards to `mod.bazingado(self)`, and the `class_eval` lives in `bazingado`.

This one is independent of the first. A variant with a reader call on the parameter — so the shape check *passes*, `replay_methods: [["bazingado", "base", "_blk"]]` — still yields `REPLAYS: []`, purely because the replay is registered under `bazingado` while `Bar` calls `bazinga`. Real `Concern` has the same split (`included` stores, `append_features` replays), so this is not incidental to the fixture.

## What the collector needs to learn

- The inverted direction: a replay method whose `class_eval` receiver is one of its own parameters and whose block comes from the enclosing scope's ivar, with the target being that parameter rather than `self`.
- One level of forwarding: `bazinga(mod)` → `mod.<method>(self)`, so the call written in the target's body reaches the method that actually replays.
- Reader-optional: keep the ivar as the join between store and replay (which is already how `BlockAnalyzer#stored_block_self_types` joins the two halves — see the note at `block_analyzer.rb:62`), rather than requiring an `attr_reader` that this direction has no use for.

Both existing spellings must keep working: example29 (two targets, two source modules) and example31 (`super` from the replayed block into an `include`d module) are the regression bar.

## Notes

- Conservatism is the point of this expander — "a chain must name one storage method, one reader, one stored block, and one constant target" — so whatever replaces the reader leg has to keep declining ambiguity just as loudly.
- Depends on #246: `example32` does not reach a fixed point without it, so its generated RBS cannot be snapshotted until that lands.

felixefelip:
Closes #247. Builds on #246 (merged), which this needed: `example32` did not reach a fixed point without it, so its RBS could not be snapshotted.

## The gap

`StoredBlockReplayExpander` knew one spelling of "keep a block, replay it on someone else": `class_eval(&param.reader)`, where `self` is the target and the source module arrives as a parameter. `ActiveSupport::Concern` is written the other way round:

```ruby
def included(base = nil, &block)
  if base.nil? then @_included_block = block else super end
end

def append_features(base)
  # …
  base.class_eval(&@_included_block) if instance_variable_defined?(:@_included_block)
end
```

`self` is the **source**; the target arrives as a parameter. `spec/dummy/app/models/example32.rb` models exactly that, and therefore got nothing — `def age`, written inside `bazingado do … end`, never landed on `Example32::Bar`.

## It is not the ivar

The obvious suspect, and the wrong one. `stored_block_ivar` picks up `@_bazingado_block = block` exactly as it picks up example29's `@bazingado_block = block`; `@storages` is populated identically for both. example29 stores in an ivar too — its `attr_reader` is just a public window onto the same slot.

Probing the collector separates the two real causes:

**1. `replay_shape` insisted the block arrive *through the parameter*.** A variant that keeps the ivar *and* adds an `attr_reader`, but reads it off `self` (`base.class_eval(&_blk)`), still yields no replay method. The axis is not "ivar vs reader" — it is *whose* state holds the block:

| | `self` inside the replay | source module | target |
|---|---|---|---|
| example29 / example31 | the **target** | the parameter | `self` |
| example32, real `Concern` | the **source** | `self` (`@_bazingado_block`) | the parameter |

Once the replaying method is inside the object that owns the slot, reading the ivar directly is the only spelling that means anything — a reader exists to let *another* object at the slot, and here there is no other object. So the `attr_reader` requirement is dropped for this direction rather than worked around.

**2. The `class_eval` must sit in the method the target's body names.** The inward direction structurally cannot put it there: the replay runs on the source with the target passed in, so the target reaches it through a forward (`bazinga(mod)` → `mod.bazingado(self)`). Independent of cause 1 — a variant with a reader call on the parameter passes the shape check and *still* declines, purely because the replay is registered under `bazingado` while `Bar` calls `bazinga`. Real `Concern` has the same split (`included` stores, `append_features` replays), so this is not incidental to the fixture.

## The change

`Collector` learns two shapes:

- `InwardReplay` — `<param>.class_eval(&@ivar)`
- `ForwardMethod` — `<param>.<callee>(self)`

and both directions now answer with the **ivar**, which is what they always had in common — and what `BlockAnalyzer#stored_block_self_types` already joins the two halves by (`block_analyzer.rb:62`). From there `resolve_replays` is shared: one method fills that slot, and the call the source wrote under its name holds the block. The outward path keeps its reader leg, now factored into `outward_ivar`.

Still deliberately conservative, which is the point of this expander:

- an inward replay's receiver must be a declared parameter, not any local in scope;
- the forward is **one hop**, not a chain — each extra link is another place a runtime value could stand in for the constant that was resolved;
- a file that somehow reads as both directions declines rather than being settled by declaration order.

## Result

```rbs
class Example32
  class Bar
    extend ::Example32::Foo

    def age: () -> Integer   # ← was absent
  end
end
```

`StoredBlockReplayImplements` emits the matching `@implements` entry from the same decision, so the sidecar and the emitted RBS cannot disagree:

```yaml
blocks:
- call: bazingado
  in: "::Example32::Baz"
  implements: "::Example32::Bar"
```

## Tests

- `spec/lib`: 1143 examples, 0 failures. Six new expander specs for the inward direction — the happy path, "needs no reader", idempotence over its own output, and three declines (no apply call, a two-hop forward, two targets for one block).
- `spec/integration`: 100 examples, **1** failure, down from 3.
- **`steep check`**: example32 raised three errors on `main` before #246 and now raises none.

  ```
  example32.rb:21:10: [warning] Method `::Example32::Baz#age` is not declared in RBS
  example32.rb:3:8:   [error] Cannot allow method body have type `(^(*untyped) -> ::Symbol | nil)` …
  example32.rb:7:8:   [error] Cannot allow method body have type `(^(*untyped) [self: singleton(::Example32::Bar)] -> untyped | nil)` …
  ```

  The last two were #246's convergence cycle; the first is this PR — `age` was emitted nowhere, so Steep saw a `def` in `Baz` that no RBS declared.

## Notes for the reviewer

- The expanded-view expectations are refreshed because this changes *which* sources the pipeline rewrites (`example32.rb` joins the set). `spec/expectations/expanded/app/models/example31.rb` comes along in the same commit: `expanded_source_spec` pins the whole set, so it cannot be added one file at a time. Its `.expanded/` sidecar was already tracked; only the expectation was missing.
- The one remaining integration failure is example31's `Cannot allow method body have type ::Integer because declared as type ::Float` at `example31.rb:26:8`. It is pre-existing on `main` and untouched here — the steep baseline still wants a conscious refresh, which I have deliberately not done since that error is a real unfixed gap rather than drift.


felixefelip:
Added `spec/expectations/models/example32.rbs` plus its `it "example32"` block in `rails_dummy_spec.rb` (6d63ef1) — the fixture had no snapshot at all, so until now the concern-direction replay was only visible in the dummy's committed `sig/rbs_infer/` output, which no spec compares against anything.

```rbs
class Example32
  @_bazingado_block: (^(*untyped) [self: singleton(Example32::Bar)] -> Symbol)?

  module Foo
    def bazinga: (singleton(::Example32::Baz) module_included) -> (^(*untyped) [self: singleton(Example32::Bar)] -> Symbol)?
    def bazingado: (?singleton(Example32::Bar)? base) ?{ (*untyped) [self: singleton(Example32::Bar)] -> Symbol } -> (^(*untyped) [self: singleton(Example32::Bar)] -> Symbol)?
  end

  module Baz
    extend ::Example32::Foo
  end
end

class Example32
  class Bar
    extend ::Example32::Foo

    def age: () -> Integer
  end
end
```

It holds down one thing per PR that touched this file:

- **#248** — `age` on `Bar`, not on `Baz` where it is lexically written, with no `attr_reader` anywhere in the fixture.
- **#246** — `[self: singleton(Example32::Bar)]` on *both* the block clause and the returned type. Those two used to alternate, and nothing else in the suite would catch a regression back into it: the convergence warning goes to stderr and fails no test.

Verified it bites rather than passing vacuously — flipping the expectation's `age` to `-> String` fails with the expected diff.

`spec/integration` is now 101 examples, 1 failure — still only example31's pre-existing `Integer`/`Float` steep error.

PR #253, opened 2026-08-16
Read a replay forwarded to each of several modules

felixefelip:
## The gap

`example36` spells the same DSL as example35, but forwards inside an iteration over a splat:

```ruby
def self.bazinga(*modules)
  modules.reverse_each do |mod|
    mod.bazingado(self)     # ← receiver is a BLOCK parameter
  end
end
```

The expander produced nothing for it, so `age` never reached `Example36::Bar`.

## The measurement

Dumping the collector's state beside example35's — every field identical except one:

| | example35 | example36 |
|---|---|---|
| `@superclasses` | `[Baz->Foo, Bar->Baz]` | same ✅ |
| `@storages` | `Foo#bazingado -> @_bazingado_block` | same ✅ |
| `@inward_replays` | `Foo#bazingado(base) &@_bazingado_block` | same ✅ |
| `@stored_calls` | `Baz.bazingado { def age }` | same ✅ |
| `@apply_calls` | `Bar.bazinga(::Baz)` | same ✅ |
| `@forwards` | `Foo#bazinga(module_included) -> bazingado` | **`[]`** ❌ |
| resolved | 1 | **0** |

The inheritance join from #251 works here. The **forward hop** is what failed.

## Cause 1 — the receiver was restricted to a parameter *name*

`forward_shape` required the receiver of `<recv>.<callee>(self)` to be one of the method's own parameters. Here it is `mod`, while `parameter_names` answers `["modules"]`.

Re-running the matcher with that one name added is what pinned it — everything else about the shape already matched:

```
method parameter_names:              ["modules"]
forward_shape(method params only):   nil
forward_shape(+ "mod"):              ["mod", "bazingado"]
```

`mod` is not arbitrary state, which is the thing the restriction exists to decline. A rest parameter holds the objects the caller passed, and iterating it yields exactly those objects — `mod` is as much "something we were handed" as `modules` is.

`handed_names` now answers that question: the method's parameters, plus the first block parameter of an element-yielding iteration over one of them, to a fixed point because the relation is transitive (an element of an element still came from the caller).

Element-yielding is an explicit list rather than "any method taking a block", because the claim being made is specifically *this name holds an element of that collection*: `hash.each { |k, v| }` hands a key first, and `obj.instance_eval { }` hands nothing at all. Only the first block parameter is read, since `each_with_index` and `each_with_object` yield a second value that is not an element.

This is not a fixture-only spelling — the generated `steep_ruby_runtime/module.rb` models `Module#include` as exactly `modules.reverse_each { |mod| mod.send(:included, self) }`.

## Cause 2 — only single-argument calls were read

`apply(A, B)` asks for A's block **and** B's, which is what a `*modules` forward means at runtime. The collector recorded an `ApplyCall` only when the call had exactly one argument, so the plural form resolved nothing.

It now emits one per argument, each resolving or declining on its own evidence. The existing one-block-per-target rule still holds, because each module carries its own block:

```ruby
class Target < Base
  apply(First, Second)
end

# →
class Wrap::Target
  def installed_first; 1; end
end

class Wrap::Target
  def installed_second; 2; end
end
```

## Effect

`Example36::Bar` gains `def age: () -> Integer`; `call_age` goes `-> untyped` → `-> Integer`; and the `not_nil self.age` precondition Steep needed while `age` was unresolvable disappears from `.steep_contracts.yml`.

The steep baseline is **unchanged** — no diagnostic added or lost. Fixed point reached after two regeneration passes, and no other dummy file moved.

## Tests

Eight specs under `a DSL forwarding to each of several modules`:

- every named module's block reaches the target (the multi-argument case)
- a single argument still resolves through the same iteration
- `each` / `map` / `select` / `each_with_index` are read too
- idempotence on a second pass
- **declines** `instance_eval`, which yields no element
- **declines** a receiver that is neither a parameter nor an element of one (`Registry.all.reverse_each`)
- **declines** a block that destructures what it is yielded
- resolves an element of an element (nested iteration)

Full suite: **1294 examples, 0 failures.**


felixefelip:
Replaced `ELEMENT_ITERATIONS` — pushed as a follow-up commit on this branch.

## Why the allowlist was the wrong shape of answer

**It wasn't even true.** "The first block parameter is the element" holds for `each`, but `Enumerable#inject` yields the *memo* first and `each_with_index` yields an *index* second. Any first-parameter rule is wrong for at least one method the list would have to contain — and a list can only ever be the methods someone happened to think of.

**And element-hood was never the question.** `ForwardMethod#parameter` is recorded and then never read again: `inward_ivar` consults only the `callee`, and the source subject comes from the apply call's constant argument. The receiver restriction is a guard against replaying onto *arbitrary state* — so what it needs is **provenance**: did this object reach us from the caller, or did we fetch it from somewhere else in the program?

## The rule now

> the method's parameters, plus the parameters of a block passed to a call **on** one of those names.

A value yielded by a method of a handed object answers the provenance question the same way the parameter does — whichever method did the yielding, and whichever position it arrived in. That claim is universally true, so it needs no list, no first-parameter special case, and no knowledge of the receiver's type (which this source-only pass could not obtain anyway).

It's also strictly *more* correct than the allowlist: `inject` and `each_with_index` now resolve, where the old first-parameter rule would have been wrong for them.

## Simpler, too

`BlockParametersNode#parameters` **is** a `ParametersNode`, so the existing `parameter_names` reads a block's bindings exactly as it already read a method's — one reader, both constructs. `first_block_parameter` is deleted. A destructuring target still declines, but now because `MultiTargetNode` carries no name, not by a rule of its own.

## Specs

- `does not depend on which method does the yielding` — now covers `each`, `map`, `select`, `each_with_index`, `reverse_each`, `each_entry`, `tap`
- new: `reads a name bound anywhere in the block's parameters` (`{ |index, mod| ... }`)
- dropped `declines an iteration that does not yield elements` — `modules.instance_eval { |mod| ... }` yields the receiver itself, so resolving it is correct; that spec was encoding a distinction that doesn't exist

**No emitted RBS changes**: example36 resolves identically, no expectation moved. 1294 examples, 0 failures.

PR #255, opened 2026-08-16
Read a replay dispatched through `send`

felixefelip:
> Stacked on #254 — targets `fix/nested-module-visibility`. Retarget to `main` once that merges.

## The gap

example38 spells the same DSL as example37, with the storage method private and the forward written through `send`:

```ruby
def bazinga(*modules)
  modules.reverse_each do |mod|
    mod.send(:bazingado, self)
  end
end

private

def bazingado(base = nil, &block)
```

The expander produced nothing for it. Every collector field matches example37 except `@forwards`, which was empty.

## It is the `send`, not the `private`

Worth stating plainly, because the two travel together and only one of them matters. A 2×2 over the two differences:

| | direct call | `mod.send(:bazingado, self)` |
|---|---|---|
| **public** | `forwards=1` ✅ | `forwards=0` ❌ |
| **private** | `forwards=1` ✅ | `forwards=0` ❌ |

Visibility is not something this pass reads — the only `private` in `collector.rb` is its own. And `bazingado do … end` in `Baz`'s class body is collected either way, since an implicit-receiver call to a private method is legal Ruby.

## Why it failed

`forward_shape` matches `<recv>.<callee>(self)` — exactly one argument, and it must be `self`. The two spellings parse differently:

```
mod.bazingado(self)          name: :bazingado   args: [SelfNode]               arity 1
mod.send(:bazingado, self)   name: :send        args: [SymbolNode, SelfNode]   arity 2
```

So it failed the arity check. And even had it passed, the callee read off `node.name` would have been `"send"` — the real callee having moved into the first argument.

## The change

A `dispatched(node)` reader answers *"what does this call actually dispatch, and with what arguments"*, unwrapping `send` / `public_send` / `__send__` when the first argument is a literal symbol or string. `x.send(:foo, a)` is statically the same call as `x.foo(a)`; the indirection is a spelling, not a data-flow question.

A **computed** name (`x.send(meth, a)`) answers `nil` — that is the arbitrary-dispatch case this pass exists to decline, and it is the same conservatism already applied to receivers, now applied to callees.

All three shape matchers go through it, so *"what method is this call"* has one answer: `forward_shape`, `inward_replay_shape`, and `replay_shape` (both its evaluator and its reader half). Only the class-body collector is deliberately left alone — an implicit-receiver call already reaches a private method, so `send` has no reason to appear there.

This is also the spelling the generated `Module#include` pseudo-code uses, for the same reason (`included` is private on `Module`):

```ruby
mod.send(:append_features, self)
mod.send(:included, self)
```

## Effect

```rbs
 class Bar
   extend ::Example38::Foo

-  def call_age: () -> untyped
+  def call_age: () -> Integer
+  def age: () -> Integer
 end
```

The baseline **loses both** entries example38 was carrying, and the `Example38::Bar#call_age` `not_nil self.age` contract goes with them:

- `example38.rb:25` — warning `Method ::Example38::Baz#age is not declared in RBS`
- `example38.rb:37` — error `Type ::Example38::Bar does not have method age`

Nothing added. Fixed point after two regeneration passes; no other dummy file moved.

## Tests

Eight specs under `a replay dispatched through \`send\``:

- reads through `send` to the method it dispatches
- reads `public_send` and `__send__` the same way
- reads a string name as well as a symbol
- **does not depend on the method being private** (the public spelling resolves identically)
- idempotence on a second pass
- **declines** a `send` whose method name is computed
- **declines** a `send` handing over something other than `self`
- reads a `class_eval` itself dispatched through `send` (`base.send(:class_eval, &@body)`)

Full suite: **1305 examples, 0 failures.**


felixefelip:
Good catch — replaced. Pushed as a follow-up commit on this branch.

`SEND_METHODS` plus a hand-rolled tuple reader duplicated `Inference::SendCall`, which has been *the one place that knows how to read a literal `send`* since #205. That's the miss `docs/engineering/prefer-principled-over-shortcuts.md` is about, and it left a second place for the same question to be answered differently later.

`SendCall.desugar` is also the better answer, not merely the pre-existing one. It returns a real `Prism::CallNode` — the same node under the dispatched name, with the name literal dropped from the arguments — so the matchers read `name`, `arguments`, `receiver` and `block` off it exactly as they read them off the original:

```ruby
def dispatched(node)
  RbsInfer::Inference::SendCall.desugar(node) || node
end
```

instead of destructuring a `[name, arguments]` tuple this file invented. `|| node` is the spelling `IntraClassCallAnalyzer` and `NewCallCollector` already use.

Declining a computed name now falls out of the design rather than needing its own branch: `desugar` answers nil, the caller keeps the original node, and the matchers see `send` with the name still sitting in the argument list — so no shape fits. The `declines a send whose method name is computed` spec still passes, unchanged.

It also picks up correctness I did not have: `desugar` handles `InterpolatedSymbolNode` (`:"a#{b}"`) explicitly as a non-literal, and it drops the name argument by rebuilding the node rather than by my slicing.

**Pure refactor** — the dummy regenerates byte-identically and no expectation moves, which is the check that this changed the code and not the behaviour. 1305 examples, 0 failures.

Issue #156, opened 2026-08-02
Generated types can outlive their evidence: the previous run's output is trusted as inference


felixefelip:
## What happens

Generation reads the RBS from the previous run as known types. That is deliberate and necessary — it is the bootstrap that lets a chain converge over two passes, and much of the pipeline depends on it.

It also means a type **survives after the evidence for it is gone**, and nothing detects that. The generated file keeps asserting something the sources no longer support, and each run re-supplies it to the next.

## Two cases, both found in the dummy

**A wrong type that pinned itself.** `Current.with` declared `-> nil` while its body returned `ProfileFormatter?`. Two full generations did not converge, because `ReturnTypeResolver#improve_method_return_types` resolves against known types *before* the Steep pass, and the previous output was among them:

```
[trace] with := … ?{ () -> untyped } -> nil
  return_type_resolver.rb:56  in improve_method_return_types
```

Deleting the file and regenerating produced the consistent `-> ProfileFormatter?` (felixefelip/rbs_infer#155).

**A right type nothing can derive any more.** The same regeneration lost `author_name: String?` from that sidecar, and it does not come back — verified with two further full runs. `app/models/current.rbs` does not carry it either, so its origin is lost to history: some earlier pipeline could derive it, and it has been living off its own output ever since.

## Why it matters

- A generated file that cannot be reproduced from its sources is not really generated — it is partly hand-state, and nobody knows which parts.
- A wrong value is *stickier* than a right one: once written, it is fed back as evidence for itself, and the pass that could correct it (Steep) never runs, because the member is no longer `untyped`.
- We do not know the size of it. `author_name` surfaced only because an unrelated change forced one file to be regenerated from scratch.

## What would answer it

Measure first: delete `spec/dummy/sig/rbs_infer/` entirely, regenerate, and diff against the committed output. Everything that fails to come back is a type living off its own output. That is a few generation rounds and gives the real dimension instead of two anecdotes.

Then the design question, which the measurement should inform:

- Should the prior output be trusted as evidence at all, or only as a starting point that a later pass may overwrite? Today `improve_method_return_types` treats a prior value as settled, so Steep never gets to correct it.
- Is there a cheap invariant — regenerate a file twice from empty and require the same result — worth running in CI for the dummy?

No change proposed here. The bootstrap earns its keep; what is missing is knowing where it stops being a bootstrap and starts being a memory.

PR #258, opened 2026-08-16
Resolve a replay whose DSL is not written where it is used

### LINKED ISSUE #256: `StoredBlockReplayExpander` can't read a replay whose applier lives on `Module` — three gaps behind `example39`
`spec/dummy/app/models/example39.rb` writes the replay the way `ActiveSupport::Concern` actually writes it: the applier is a method on `Module` (core), and it dispatches to a method the *argument* owns. `StoredBlockReplayExpander` resolves nothing, so `age` never reaches `Example39::Bar`.

```ruby
# spec/dummy/lib/rails_ext/module.rb
class Module
  def banana(*modules)
    modules.reverse_each { |mod| mod.send(:bananed, self) }
    self
  end

  private

  def bananed; end
end
```

```ruby
# spec/dummy/app/models/example39.rb
module Example39
  module Foo
    private

    def bananed(base = nil, &block)
      if base.nil?
        @_bananed_block = block
      else
        base.class_eval(&@_bananed_block) if @_bananed_block
      end
    end
  end

  module Baz
    extend Example39::Foo

    bananed do
      def age = 31
    end
  end

  class Bar
    extend Example39::Foo

    banana(Example39::Baz)

    def call_age = age + 10
  end
end
```

At runtime this works — `Example39::Bar.new.call_age # => 41` — once `module.rb` is actually loaded, which needed `config/initializers/rails_ext.rb` (`lib/rails_ext` is excluded from `autoload_lib`, since those files reopen framework classes and don't name a Zeitwerk-autoloadable constant).

## Measured

Dumping every `Collector` field for #253's `example38` (which resolves) beside `example39` (which doesn't), exactly **one** differs:

```
example38  forwards  [ForwardMethod(owner="Example38::Foo", method="bazinga", parameter="mod", callee="bazingado")]
example39  forwards  []
```

`storages`, `inward_replays`, `stored_calls`, `apply_calls` and `providers` all have the same shape. The empty `forwards` is what declines the join: in `resolve_replays`, both `outward_ivar("Example39::Foo", "banana")` and `inward_ivar("Example39::Foo", "banana")` answer nil, `ivars` is empty, and the candidate is dropped.

`forwards` is empty because `banana` is not in this file.

## That is not the whole story

Three placements of the same applier, everything else held fixed:

| variant | result |
|---|---|
| as written (applier in `lib/rails_ext/module.rb`) | no expansion |
| applier inlined into `Example39::Foo` | `def age` emitted |
| applier inlined as a `class Module` reopen, same file | **no expansion** |

So the file boundary is one cause of three, and each of the following is independently sufficient to decline.

### 1. The seam is single-file

`StoredBlockReplayExpander.expand(source)` takes only the source string. There is no corpus, so a chain whose links are written in different files cannot be seen however the rest is fixed.

### 2. A `Module`/`Class` reopen is not a DSL provider

Concatenating `module.rb` ahead of `example39.rb` gets the forward collected, with the wrong owner:

```
forwards    [ForwardMethod(owner="Module", method="banana", parameter="mod", callee="bananed")]
providers   {"Example39::Foo" => #<Set: {"Example39::Baz", "Example39::Bar"}>}
=> replays: []
```

`resolve_replays` only iterates owners that are keys of `providers`, and `dsl_providers` knows two relations: `extend X`, and the superclass chain declared in the file. `Bar` can call `banana` because it *is* a `Module` instance — a relation `dsl_providers` has no way to express. `Module` is never consulted, and the forward is orphaned.

### 3. The forward and the replay must share an owner

Isolated with no `Module` involved at all — single file, applier in an ordinary `module Applier` that only `Bar` extends, storage/replay in `Foo` that only `Baz` extends. **No expansion.**

The constraint is structural, in `inward_ivar`:

```ruby
forwards = @forwards.select { |forward| forward.owner == owner && forward.method == method }
replays  = @inward_replays.select { |replay| replay.owner == owner && replay.method == forwards.first.callee }
```

Both must sit in the same owner. `resolve_replays` adds the matching requirement one level up — `apply.subject` and `source_subject` must both be subjects of the *same* provider.

In `example38` that holds by accident: `bazinga` and `bazingado` are both in `Example38::Foo`. In `example39` it does not. `banana` belongs to `Module` (reaching `Bar`), `bananed` to `Foo` (reaching `Baz`). What joins them is the **receiver of the dispatch** — `mod` is `Baz`, and it is `Baz` that extends `Foo` — not the forward's owner.

This is what `ActiveSupport::Concern` does: `Module#include` forwards to `mod.append_features(self)`, a method on the concern itself. Two owners, joined only through the argument.

## Fixing it

All three, together — the measurements above show no one of them is sufficient:

1. give the expander access to the corpus rather than one source string;
2. treat a `Module`/`Class` reopen as a provider for every declaration;
3. resolve a forward's callee against the provider of its **argument**, not against the forward's own owner.

(3) is the one that carries meaning beyond this fixture: it is what makes the join follow Ruby's dispatch instead of lexical co-location, and (1) and (2) mostly exist to put the relevant `def`s in front of it.

## Related

`example40` writes a fourth variation — `Foo#bazingado` delegates to a memoized `Example40::Bazinga` instance, so the storage and the replay live in a class nobody extends, reached only through an ivar. That needs the value of the ivar followed, which is data-flow rather than lexical, and is out of scope here.

felixefelip:
Closes #256.

`example39` writes the replay the way `ActiveSupport::Concern` writes it: the applier is `Module#include` — a method of a class the project reopens, in a file that has nothing else to do with the fixture — and it reaches for a method the **argument** owns.

```ruby
# spec/dummy/lib/rails_ext/module.rb
class Module
  def banana(*modules)
    modules.reverse_each { |mod| mod.send(:bananed, self) }
  end
end
```

Three things declined it, each on its own. All three had to give, which is what #256 measured and what this does.

## 1. The seam was single-file

`expand` took a source string and nothing else, so a chain with links in two files could not be seen however the rest was fixed.

It now takes `sources:` — a new `Project::ConstantSources` answering which files declare *or reopen* a constant. Candidates come from `FileIndex` **and** `SourceIndex`, and every one is parsed before it counts: for a core reopening the path convention is no guide at all, since `class Module` can be written anywhere. Verification is what makes the over-broad candidate list harmless — a file that merely mentions `Module` in a comment fails the walk.

Only method **shapes** are absorbed from those files. An `apply` written elsewhere would name a target this rewrite cannot reach, and a spec pins that.

## 2. A core reopening was not a provider

`dsl_providers` knew two relations: `extend X`, and the superclass chain. `Bar` can call `banana` because it **is** a `Module` — neither relation expresses that, so with both files concatenated into one string the forward was still orphaned:

```
forwards    ForwardMethod(owner="Module", method="banana", callee="bananed")
providers   {"Example39::Foo" => {Baz, Bar}}
=> replays: []
```

`Module` now provides every declaration, `Class` every class. Costless where nothing reopens them: an owner with no collected shapes answers no slot, so the join declines exactly where it declined before.

## 3. The two halves had to share an owner

The applier is dispatched on the **subject** (`banana` arrives in `Bar`'s body) and the callee it forwards to on the **argument** (`mod.send(:bananed, self)` runs on `Baz`). Those are supplied by whatever supplies each, and they need not be the same module — `banana` belongs to `Module`, `bananed` to `Foo`. `resolve_replays` asks the two questions separately now.

This one is isolated by a spec with no other file and no `Module` involved at all: an applier in one module, a keeper in another, joined only through the argument.

## Threading

`sources:` is **required**, with `ConstantSources::NONE` as the explicit way to say "no project". A caller that forgot it would resolve strictly less and say nothing about it — the silent-wrong case in `docs/engineering/required-threaded-deps.md`.

## Two readers widened to keep one answer

- **`ModuleSelfTypeGenerator`** built its corpus from the files Steep *checks*. A Steepfile `ignore` says "report no diagnostics here", not "this file is not part of the program" — and `lib/rails_ext/**` is ignored precisely because stock Steep cannot impl-check a raw framework reopening (#38). Reading the narrow list would let the `@implements` sidecar resolve fewer replays than the expander does, and the two disagreeing about which class a block defines its methods on is the one thing `StoredBlockReplayImplements` exists to prevent. Caught by measurement: the sidecar had no `blocks:` entry for example39 while the RBS already had `age`.
- **`expanded_source_spec`** passed `app/ + sig/` where the Makefile passes `app/ lib/ sig/`, so it was asserting an expanded view the pipeline never produces — which is also why it went green on a change it should have caught. Widening it added exactly one file to the set and moved no existing snapshot.

## A limit, stated

`expand` still gates on the source naming `class_eval`/`module_eval` before it parses anything, which is what keeps the pass off every file in the project. A DSL that leaves *no* trace in the file using it is not examined. Every shape here still shows the replay half in the file, which is how the fixtures are written; a spec pins the limit rather than leaving it implicit.

## Result

```
- def call_age: () -> untyped
+ def call_age: () -> Integer
+ def age: () -> Integer
```

The `blocks:` entry appears in the self-types sidecar, the `Example39::Bar#call_age` not-nil contract disappears (there is nothing left to require), and both baseline errors are gone — the two lines added in #257 come straight back out:

```
- app/models/example39.rb:19:10: [warning] Method `::Example39::Baz#age` is not declared in RBS
- app/models/example39.rb:31:6: [error] Type `::Example39::Bar` does not have method `age`
```

21 new specs (7 for `ConstantSources`, 14 for the join); 1333 in the suite, 0 failures. `example38` and `example40` regenerate unchanged.


felixefelip:
## `CORE_REOPENS` is now derived, not asserted (61c34b7)

Following a review question — could this come from the ancestors Steep already knows, instead of a hand-written list?

**The Steep query does answer.** `RbsDefinitionResolver#method_owner` already walks RBS's singleton ancestor graph:

```
singleton(::Example39::Bar)  banana   -> "::Module"
singleton(::Example39::Baz)  bananed  -> "::Example39::Foo"
```

The second line means it would have replaced the `extend` half of `dsl_providers` too, not just the core case.

**But it answers only because rbs_infer already wrote the signature.** Moving `sig/rbs_infer/lib/rails_ext/module.rbs` aside and re-running:

```
singleton(::Example39::Bar)  banana   -> nil
```

The expander runs *before* the parse and decides what gets generated. Reading generated RBS there makes the expansion depend on the previous run's output — the circularity of #156 — and makes the first pass over a cold checkout expand nothing. The `Collector`'s own first line is "It keeps syntax, not guessed types".

**So it is derived from Ruby instead**, which needs no environment and cannot be stale: the superclass chain of `Class` and of `Module` — what `self` in each kind of body actually is.

```
"class"  => [Class, Module, Object, BasicObject]
"module" => [Module, Object, BasicObject]
```

That fixes a real gap the hand-written pair had: `class Object; def banana` makes the applier callable in every class body just as surely, and `%w[Module Class]` missed it. There is a spec for it now.

**Not `ancestors`, though it also derives.** `ancestors` reads the live process rather than the language — measured, `Class.ancestors` answers `PP::ObjectMixin` and `JSON::GeneratorMethods` here, because `pp` and `json` inject into `Object` and rbs_infer loads them. Which gems the *analyzer* requires is no fact about the analyzed project, and it would make the list differ between environments. A superclass chain cannot be injected into. `Kernel` is left out knowingly: it is a module, so nothing at runtime tells it from the two injected ones.

**Cost**, since going from 2 names to 4 was the open question — measured over the dummy's 141 files:

```
Class        candidates=1  declaring=0   0.2 ms
Module       candidates=4  declaring=2   0.5 ms
Object       candidates=1  declaring=0   0.1 ms
BasicObject  candidates=1  declaring=0   0.2 ms
TOTAL                                    1.0 ms   (once per run, memoized)
WARM                                     0.002 ms (per file expanded after the first)
```

My worry that `files_referencing("Object")` would sweep hundreds of files was wrong: `SourceIndex` keys on CamelCase tokens, and `Object` rarely appears as a bare constant in app code.

The dummy regenerates **byte-identically** and no expectation moves, which is the check that this changed the derivation and not the behaviour. 1336 examples, 0 failures.

PR #286, opened 2026-08-28
Don't count a `return` the checker has already refuted

felixefelip:
## The bug

`ReturnTypeResolver#has_nil_return?` was a syntactic walk over the body — any bare `return` (or `return nil`) anywhere in it widened the method's return type to `T?`, whatever the branch it sits in was typed:

```ruby
def has_nil_return?(defn)
  RbsInfer::Analyzer.find_all_nodes(defn) do |node|
    next false unless node.is_a?(Prism::ReturnNode)
    node.arguments.nil? || node.arguments.arguments.any? { |arg| arg.is_a?(Prism::NilNode) }
  end.any?
end
```

Nothing asks whether that `return` can run. So a guard written against a value that is already non-nil — the shape every `before_action`-established reader has — turned a method that never returns nil into a nilable one:

```ruby
def pessoa_ativa
  return if current_user.nil?   # dead: current_user is non-nil past authenticate_user!

  current_user.pessoas.find_by(id: session[:pessoa_id]) || current_user.pessoa_titular
end
```

(`find_by` gives `(Pessoa & Validated)?` and `pessoa_titular` is non-nil under `User::Validated` — `validates :pessoa_titular, presence: true`. The **only** `nil` in that method was the dead `return`.)

And the type is not where it stops. The caller does `Current.pessoa = pessoa_ativa`; with a nilable argument the `unless value.nil?` inside `Current#pessoa=` stays live, so `Current.caderneta` was never established, the action's `Current.caderneta.vacinas` was a `NoMethod`, the ivar write typed as `"nil"`, and `infer_ivar_types` dropped the declaration outright. The ivar stopped being generated three steps from the guard, with nothing in between pointing back at it.

## The fix

Steep already had the answer and was never asked. It decides reachability while typing the `if` and reports the dead clause as `UnreachableBranch` — verified on the real file:

```
UnreachableBranch L13: if
send L13 current_user       => (::User & ::User::Validated)
send L13 current_user.nil?  => <% Logic::ReceiverIsNil %>
```

It covers every spelling I tested: `return if x.nil?`, a local holding the same value, `return unless x`, and a literal-false condition.

`SteepBridge#unreachable_branch_ranges` reads that verdict back, and `has_nil_return?` skips any `return` inside one — the widening now follows the checker instead of the parser. The ranges are in **byte** offsets: Steep's parser indexes its buffer by character, so one non-ASCII byte earlier in the file would slide every range onto the wrong node.

`ClassMemberCollector` has a second copy of `has_nil_return?`. It stays syntactic and is commented as such: that pass runs before any type-check, so there is no verdict to consult; its guess is taken back by the resolver's narrowing pass, which does have the bridge.

## What moves

`posts_controller`'s `build_filtering_for_current_user` narrows to `-> PostFiltering` for the same reason — its one caller assigns `Current.user` immediately before, so `return unless Current.user` is dead by the closed-world reading the rest of the pipeline already uses. Steep flags it itself; I checked before accepting it.

`example56` is the regression fixture, with a live guard as the control so the fix cannot over-fire:

| | before | after |
|---|---|---|
| `label` (dead `.nil?` guard) | `String?` | **`String`** |
| `shout` (dead `unless` guard) | `String?` | **`String`** |
| `maybe_label` (**live** guard) | `String?` | `String?` |
| `self.call` (cascade) | `untyped` | **`Integer`** |

## Baseline

Three diagnostics enter `steep_baseline.txt` and belong there:

```
app/controllers/posts_controller.rb:76:4: [error] The method cannot return a value of type `nil` because declared as type `::PostFiltering`
app/models/example56.rb:29:6: [error] The method cannot return a value of type `nil` because declared as type `::String`
app/models/example56.rb:36:6: [error] The method cannot return a value of type `nil` because declared as type `::String`
```

Same asymmetry, the other side of the fence: having declared the branch unreachable, Steep still checks its `return` against the declared return type. `TypeConstruction`'s `when :return` reports `ReturnTypeMismatch` with no reachability guard, even though `truthy.unreachable` was computed a few lines earlier in the `:if` handler. That is a defect in the fork, not in this change, and its fix erases all three entries.

## Tests

`make test` green (1353 + 123, 0 failures). Reverting only `lib/` makes the `example56` snapshot fail exactly on the two dead guards, with the live one unchanged.

Verified end-to-end on the app the report came from: `pessoa_ativa` becomes non-nil, `set_authenticated_user` establishes `Current.caderneta`, the entry facts carry it into `CadernetaController#index`, and `@vacinas` is generated again.

https://claude.ai/code/session_012BrWNAbYgzR7CKVuoKk6DS

### DIFF
--- a/lib/rbs_infer/inference/class_member_collector.rb
+++ b/lib/rbs_infer/inference/class_member_collector.rb
@@ -571,6 +571,11 @@ module RbsInfer::Inference
       type
     end
 
+    # Purely syntactic, unlike `ReturnTypeResolver`'s: this pass runs BEFORE any
+    # type-check, so there is no reachability verdict to consult yet. A `return`
+    # the checker will later prove dead is counted here and taken back by the
+    # resolver's narrowing pass, which does have the bridge
+    # (felixefelip/rbs_infer#286).
     def has_nil_return?(defn)
       RbsInfer::Analyzer.find_all_nodes(defn) do |node|
         next false unless node.is_a?(Prism::ReturnNode)
--- a/lib/rbs_infer/inference/return_type_resolver.rb
+++ b/lib/rbs_infer/inference/return_type_resolver.rb
@@ -99,7 +99,7 @@ module RbsInfer::Inference
               steep_type = "self" if self_return?(m, steep_type, self_types)
 
               # Check for early return nil in body
-              if defn && has_nil_return?(defn)
+              if defn && has_nil_return?(defn, dead_ranges: dead_ranges(parsed_target))
                 steep_type = RbsInfer::Signatures::RbsParserUtil.nilablize(steep_type)
               end
 
@@ -142,7 +142,7 @@ module RbsInfer::Inference
             next unless RbsInfer::Signatures::RbsParserUtil.nilablize(steep_type) == current_type
 
             defn = def_map[m.name]
-            next if defn && has_nil_return?(defn)
+            next if defn && has_nil_return?(defn, dead_ranges: dead_ranges(parsed_target))
 
             steep_type = "self" if self_return?(m, steep_type, self_types)
             m.signature = m.signature.sub(/-> #{Regexp.escape(current_type)}$/, "-> #{steep_type}")
@@ -163,7 +163,7 @@ module RbsInfer::Inference
             steep_type = "self" if self_return?(m, steep_type, self_types)
 
             defn = def_map[m.name]
-            if defn && has_nil_return?(defn)
+            if defn && has_nil_return?(defn, dead_ranges: dead_ranges(parsed_target))
               steep_type = RbsInfer::Signatures::RbsParserUtil.nilablize(steep_type)
             end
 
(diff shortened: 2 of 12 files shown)


felixefelip:
Added unit coverage — the snapshot pins what the pipeline emits, but says nothing about the two pieces underneath, and one of them it structurally *cannot* reach.

**`ReturnTypeResolver#has_nil_return?`** (5 examples) — the body has one dead `return` and one live one, so the filter is exercised per node rather than for the method as a whole:

| dead_ranges | result |
|---|---|
| `[]` | counts (true) |
| a range covering neither | counts (true) |
| covering only the first `return` | still counts — the live one remains (true) |
| covering both | counts nothing (false) |

**`SteepBridge#unreachable_branch_ranges`** (3 examples) — the refuted guard, the live guard, and the case **no fixture in this repo can produce**: a multibyte character before the guard. Steep's parser indexes its buffer by character and Prism reads bytes, so on an ASCII file the two agree and a character-indexed implementation passes every existing test. With `# ação, coração, não` on the line above:

```ruby
code.byteslice(range.begin, range.size)  # => "return"   ✓
code[range.begin, range.size]            # => "if nam"   ← what characters would have given
```

The spec asserts both, so a regression reads as "it went back to characters" rather than an offset nobody can place.

Both were checked against the mutation they describe:

- reverting the `dead_ranges` filter → fails `counts nothing once every return is proved dead`, and only that;
- making `byte_offset` return its argument → fails `answers in byte offsets, not character offsets`, and only that.

`make test` green: 1361 + 123, 0 failures.

PR #294, opened 2026-08-30
Give a delegated method the parameters of the method it delegates to

felixefelip:
`delegate :default_value?, to: :class` in `Filter::Fields` came out as

```rbs
def default_value?: () -> bool
```

for a method that takes `(key, value)` — a signature its own call sites contradict.

## Why

`resolve_delegate_methods` resolved the **return** type and wrote `() ->` in front of it. The parameters were never asked about, for any delegate. It stayed invisible because every `delegate` in the dummy forwards a zero-argument reader, where `()` is the right answer.

`to: :class` was a second defect on the same line. The receiver was derived by capitalizing the reader's name — and capitalizing `"class"` gives the constant `Class`, so every lookup went to `::Class`, which has none of these methods. The return type that did come out was an accident: a same-named member elsewhere in the file filled the `untyped` in.

## The rule

A delegated method accepts what the method it forwards to accepts.

- **Parameters** are copied from the target's declaration, one overload per declared overload. Only the RBS can answer: the parameters live in the declaration, and *building* the definition is what reaches a method the target holds through a superclass, an `include` or an `extend`. A run whose target has no RBS yet keeps `()` and gets the parameters on the pass after — the same convergence every cross-file type here depends on.
- **The return** keeps its own resolution, which reads more than a declaration does (intersections, nilable receivers, source-only classes) and already handles `allow_nil`.
- **`to: :class`** forwards to `self.class`, so the receiver is a **singleton**: the target's own, or — inside a concern — the hosts its `@type instance:` annotation names, which is where a `class_methods do` block actually lands.

## Result

```rbs
-  def default_value?: () -> bool
+  def default_value?: (Symbol key, untyped value) -> bool
```

## Pins

- `example60` — both halves. `Example60#stamp` delegates to a reader, and nothing in the fixture states its parameters: `Example60Printer#stamp` is itself typed from a call site, and the delegate copies what the previous pass wrote down. `Example60::Labels` is a concern delegating `to: :class` into its own `class_methods do` block. Without the change both come out `() -> String`.
- `delegate_signature_spec.rb` — the parameter copy, `to: :class` on a plain class and inside a concern, an overloaded target, `allow_nil` nilablizing only the return, and the fallback that keeps `()` when no declaration answers.

## Also in this branch

One line of `spec/dummy/sig/rbs_infer/sig/generated/steep_ruby_runtime/module.rbs`: `singleton(::Example59::Naming)` was one regeneration pass behind when #293 was committed.

Full suite green (1526 examples), dummy Steep baseline unmoved, and no other generated RBS in the dummy moves.

## Not in this branch

`delegate` is read by a `when :delegate` in `ClassMemberCollector` — core code naming an ActiveSupport DSL, which `docs/engineering/keep-core-framework-agnostic.md` says should be an extension. It cannot be a `SourceExpander`, though: the desugaring that produces the right RBS needs the target's declared parameters, and an expander runs before the RBS environment exists. Moving it behind a seam is worth doing separately; this PR leaves it where it is.

https://claude.ai/code/session_01AqWEMG99PggU3TEG3vbx74


lib/rbs_infer/signatures/rbs_definition_resolver.rb
```diff
@@ -349,6 +382,29 @@ def compute_method_owner(type_str, method_name)
...
+      )
+      rendered = sentinel.to_s.delete_suffix(" -> void")
+      # Drop the leading `::` of every name, the way every other emitted type has
+      # it dropped. Not `format_rbs_return_type`'s pattern: a parameter list puts
+      # a name after `?`, `*`, `**` and `&` too, and `?::Integer times` came out
+      # with the `::` still on it. What must survive is the `::` INSIDE a name
+      # (`Example60::Labels`), which is the one preceded by an identifier.
+      rendered.gsub(/(?<![A-Za-z0-9_:])::/, "")
+    rescue RBS::BaseError, NoMethodError
+      nil
```
felixefelip: Não é melhor pelo menos imprimir em algum lugar esse erro?

lib/rbs_infer/signatures/rbs_definition_resolver.rb
```diff
@@ -232,6 +232,39 @@ def method_owner(type_str, method_name)
...
+      type_name = build_rbs_type_name(class_name)
+      return [] unless type_name
+      return [] unless rbs_builder.env.class_decls.key?(type_name)
+
+      definition = kind == :singleton ? rbs_builder.build_singleton(type_name) : rbs_builder.build_instance(type_name)
+      method = definition.methods[method_name.to_sym]
+      return [] unless method
+
+      method.defs.filter_map { |d| render_parameters(d.type) }.uniq
+    rescue RBS::BaseError
```
felixefelip: Aqui também, não é melhor imprimir o erro, ou algo do tipo?

felixefelip: Boa — corrigido em 751afdc.

Agora avisa com classe e mensagem, no formato `[rbs_infer] ...` que o SteepBridge e o gerador do AR runtime já usam, nomeando o que estava sendo perguntado:

    warn "[rbs_infer] could not read #{kind} parameters of #{class_name}##{method_name}: #{e.class}: #{e.message}"

O motivo de valer a pena: um delegate que perde os parâmetros sai como `()`, que se lê como resposta e não como falha — e o run não tem outro lugar onde isso apareça.

lib/rbs_infer/signatures/rbs_definition_resolver.rb
felixefelip: Mesma correção em 751afdc, e aqui fui um passo além: tirei o `NoMethodError` do rescue.

    rescue RBS::BaseError => e
      warn "[rbs_infer] could not render #{method_type}: #{e.class}: #{e.message}"
      nil

Ele era defensivo contra uma API do RBS que eu não tinha confirmado (`Types::Function#update`) — já confirmei que existe. Um method type que isso não consegue imprimir é bug daqui, não um fato sobre o projeto sendo lido, e engolir viraria um delegate que só *parece* sub-tipado. `RBS::BaseError` fica, porque esse é o modo de falha legítimo, e agora fala.

Um overload que não renderiza sai da lista; os outros continuam valendo.

Rodei `make rbs_infer` no dummy: nenhum dos dois warns dispara.

PR #318, opened 2026-09-06
Substitute a forwarded block's self, which named another receiver

### LINKED ISSUE #317: A block parameter typed `self` is forwarded across a method boundary, where `self` means something else
## What is emitted

Two integration examples have been red since felixefelip/rbs_infer#284, and both fail on the same line. `spec/dummy/app/models/example55/foo.rb`:

```ruby
module Example55::Foo
  def banana_class_method(&block)
    mod = const_defined?(:BananaMethods) ? const_get(:BananaMethods) : const_set(:BananaMethods, Module.new)

    mod.module_eval(&block)
  end
end
```

```diff
- def banana_class_method: () ?{ (*untyped) -> Symbol } -> untyped   # the checked-in expectation
+ def banana_class_method: () { (self) -> Symbol } -> Symbol         # what is emitted today
```

`ActiveSupport::Concern#class_methods` in `spec/expectations/steep_ar_runtime/` has the same diff, from the same shape.

**Two changes on one line, and only one of them is right.** Regenerating the snapshots as they stand would freeze the wrong half.

## `(self)` is wrong

It is not invented — it is forwarded verbatim from core RBS:

```rbs
def module_eval: [U] () { (self mod) [self: self] -> U } -> U
```

(`class_eval` is `alias class_eval module_eval`, so it carries the same shape.)

The problem is that **`self` is not portable across a method boundary.** In core's signature `self` means *the receiver of `module_eval`* — the `mod` local. Copied into `banana_class_method`'s signature it is read as *the receiver of `banana_class_method`* — `Example55::Foo`. Different objects.

Measured at runtime:

```ruby
module Foo
  def self.banana(&block)
    mod = const_defined?(:Banana) ? const_get(:Banana) : const_set(:Banana, Module.new)
    mod.module_eval(&block)
  end
end
Foo.banana { |arg| p arg }
```
```
o bloco recebe: Foo::Banana (Module)
Foo é: Foo
```

The block receives `Foo::Banana`. The emitted signature claims it receives `Foo`.

`Module` is the honest conservative answer. Resolving it further to the actual module would be better, and is a separate question — the bug is that a `self` written in one method's signature is being read in another's.

## The required block is right

The other half of the diff — `?{` becoming `{` — is a fix, not a regression:

```
module_eval(&nil): ArgumentError: wrong number of arguments (given 0, expected 1..3)
```

`banana_class_method` without a block raises. The old `?{` said the block was optional, and it is not.

## Where

`Inference::BlockSignatureResolver` binds a block's SELF (`bind_block_self`, for `[self: …]`) but nothing rewrites a `self` appearing as a **parameter type**. A forwarded block type needs `self` replaced by what it denoted at the source — or, failing that, widened to the erased type.

## Acceptance

- A block parameter typed `self` in a forwarded signature is never emitted as `self` in a method of another owner.
- `Example55::Foo#banana_class_method` reads `() { (Module) -> Symbol } -> Symbol` — the required block kept, the parameter honest.
- `spec/expectations/models/example55/foo.rbs` and `spec/expectations/steep_ar_runtime/active_support/concern.rbs` are regenerated **after** the fix, so the snapshots record the corrected type rather than the current one.
- The dummy's Steep baseline does not regress.

## Note on how this was found

Bisecting the two failures over the 31 commits since #284 lands on `84e2ed4` ("Remove unecessary self type annotation in ApplicationHelper", two lines in `spec/dummy/app/helpers/application_helper.rb`). That is where the tree first produces the failure, **not** the cause: restoring those two lines today does not revert the output. The mechanism by which that commit flipped it is unexplained and is now beside the point — the emitted type is wrong for a reason that stands on its own.

Worth recording separately: the second failure has since grown an extra diff line (`def self.extended: (untyped base)` → `(Object base)`) introduced by #309, which was never regenerated. A test that is already red hides the next one.

https://claude.ai/code/session_0179Um6Qoek3LMYrqHKUPuos

felixefelip:
Closes #317 - and with it the two integration examples that have been red since #284.
**The suite is green again: 139 examples, 0 failures.**

## The bug

`BlockAnalyzer#callee_block_params` formatted the callee's block parameter types verbatim.
Core declares:

```rbs
def module_eval: [U] () { (self mod) [self: self] -> U } -> U
```

so `Example55::Foo#banana_class_method`, which forwards its block to `mod.module_eval`,
came out as `() { (self) -> Symbol } -> Symbol`.

**`self` is not portable.** In core's signature it names the receiver of `module_eval` -
the `mod` local. Written on `banana_class_method` it is read as *that* method's receiver,
`Example55::Foo`. Different objects, measured:

```
o bloco recebe: Foo::Banana (Module)   <- the block's actual argument
Foo e: Foo                             <- what the signature claimed
```

## The fix

`self` is substituted with what it denoted **at the call site** - the type of that send's
receiver:

    - def banana_class_method: () { (self) -> Symbol } -> Symbol
    + def banana_class_method: () { (Module) -> Symbol } -> Symbol

Two cases fall out of the substitution rather than being special-cased:

- an **implicit receiver** (`class_eval(&block)`) means the callee's `self` *is* ours, so
  the word survives unchanged and stays correct;
- a `self` **nested** in a larger type, or a receiver the checker will not name, widens to
  `*untyped` - carrying the word across would say something about the wrong object.

## Three expectation lines, three causes

| line | why |
|---|---|
| `(Module)` | this fix |
| `{` instead of `?{` | already correct in the output, wrong in the snapshot: `module_eval(&nil)` raises `ArgumentError`, so the block is not optional |
| `self.extended: (Object base)` | debt from #309, where the hook gained a real call site and only `example62.rbs` was regenerated |

None of the three freezes a bug - which is why the snapshots were regenerated *after* the
fix rather than before.

## Worth recording

The second failure had quietly grown that third line, and nobody saw it - **because the
example was already red.** A failing test hides the next one.

The bisect over the 31 commits since #284 lands on `84e2ed4` (two lines removed from
`application_helper.rb`), but restoring them does not revert the output. That is where the
tree started failing, not why. The emitted type was wrong for a reason that stands on its
own, which is what this fixes.

## Verification

- 1408 unit examples green but for the pre-existing environmental failure (rbs 4.2.0 names
  `Array`'s parameter `E`; the spec expects `Elem`).
- **139 integration examples, 0 failures.**
- Steep baseline on the dummy unchanged.

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0179Um6Qoek3LMYrqHKUPuos


lib/rbs_infer/signatures/steep_bridge/block_analyzer.rb
```diff
@@ -214,22 +214,39 @@ def required_callee_block(typing, send_node)
...
+    def receiver_type(typing, send_node)
+      node = send_node.children[0]
+      return "self" if node.nil?
+      return nil unless node.is_a?(Parser::AST::Node)
+
+      RbsInfer::Signatures::SteepBridge::TypeFormatter.format_type(typing.type_of(node: node))
+    rescue StandardError
+      nil
+    end
+
```
felixefelip: Esse rescue genérico e sem salvar o erro não é uma boa prática pois atrapalha o debug

felixefelip: Boa observacao, e o rescue nem precisava ser generico.

Steep levanta uma classe so para o caso que de fato se espera aqui (`lib/steep/typing.rb`):

```ruby
class UnknownNodeError < StandardError

def type_of(node:)   # raise UnknownNodeError.new(:type, node: node)
def call_of(node:)   # raise UnknownNodeError.new(:call, node: node)
```

Entao ficou:

```ruby
rescue Steep::Typing::UnknownNodeError
  nil
end
```

"o checker nao tem tipo para este no" continua virando `nil`; qualquer outra coisa sobe.

**Estreitei dois, nao um.** O rescue do `required_callee_block`, tres linhas acima, tem a mesma forma e chama `call_of`, que levanta a mesma classe - foi de la que copiei o padrao ao escrever o `receiver_type`. Deixar um estreito ao lado de um generico ficaria pior.

**Considerei nao usar rescue.** Steep tem `has_type?(node)`, mas o `type_of` sobe para o `parent` quando a tabela local nao tem, e o `has_type?` so olha a local - o predicado responderia diferente em typing aninhado. O rescue especifico e fiel onde o predicado aproxima.

**Sobraram dois genericos no arquivo** (linhas 186 e 197, um deles `rescue nil` de modificador), pre-existentes e fora deste hunk. Nao toquei para nao alargar o diff da revisao; e so dizer se quiser nesta branch ou num PR de limpeza.

Verificado: 1408 unitarios com so a falha ambiental de sempre; 139 de integracao, 0 falhas - estreitar rescue pode aflorar o que ele engolia, e nada aflorou.

Commit: 73c6c64

PR #319, opened 2026-09-06
Fix the two failing specs, and the shipping gap the second one pointed at

felixefelip:
**1579 examples, 0 failures** — the whole suite, in the environment the lock now pins.

## 1. `rbs_definition_resolver_spec` — a third party's spelling

The spec pinned `Array[unchecked out Elem]`; rbs 4.2.0 renamed it to `E`. The implementation only renders what RBS declares (`entry.type_params.map(&:to_s)`), so there was nothing to fix in rbs_infer.

`Hash[K, V]` did not change — and its assertion never ran, because the `Array` one failed first.

## 2. `class_methods do` in `spec/bin` — my regression, from #314

Bisected: green at `d9337c1` (#313), red at `a4f1047` (#314).

It is the exact cost I wrote into #314's own description — *"the pass now REQUIRES the runtime sidecar in the corpus"* — filed as a risk when it had already broken a real test. I did not see it because **I never ran `spec/bin/`**: three times this week I called the suite green having run 1547 of 1579 examples.

The fixture built a temp project with the AR-runtime sidecar and without the language's. It now generates both, which is what a real project has on disk.

## And the defect the spec was pointing at

Asked how a *user of the gem* would produce that sidecar, the answer was: they could not.

| | |
|---|---|
| `extensions/rails/tasks/` | 6 rake tasks |
| `ruby_runtime` | no task, not in the Railtie, not in the README — only a `Makefile` target used for the dummy |

So since #314 an application's `extend` resolves to nothing, silently, and there is no command to fix it.

Added `lib/rbs_infer/tasks/rbs_infer_ruby_runtime.rake` — under `tasks/`, **not** `extensions/rails/tasks/`, because `RubyRuntimeGenerator` is core: its own comment says `include` is plain Ruby and that is why it is not an extension. Loaded from the Railtie, listed in the README's generator table, and run for real:

```
$ rake rbs_infer:ruby_runtime:all
Generated language-runtime pseudo-code: .../sig/generated/steep_ruby_runtime
```

## Still open, separately

- `spec/dummy/sig/rbs_infer/sig/generated/steep_ruby_runtime/module.rbs` has no expectation — the only runtime sidecar without one, and since #314 the most load-bearing. It moved this week (`prepend_features: (Module mod)` → `(untyped mod)`) and nothing caught it.
- The root `Gemfile` uses `path: "../steep"`, which does not resolve from a git worktree; `spec/dummy/Gemfile` already uses an absolute path for the same gem.

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0179Um6Qoek3LMYrqHKUPuos

### DIFF
--- a/lib/rbs_infer/tasks/rbs_infer_ruby_runtime.rake
+++ b/lib/rbs_infer/tasks/rbs_infer_ruby_runtime.rake
@@ -6,8 +6,7 @@ namespace :rbs_infer do
   namespace :ruby_runtime do
     desc "Generate the language-runtime pseudo-code sidecar for Steep (sig/generated/steep_ruby_runtime/)"
     task :all do
-      app_dir = defined?(Rails) ? Rails.root.to_s : Dir.pwd
-      dir = RbsInfer::Project::RubyRuntimeGenerator.new(app_dir: app_dir).generate
+      dir = RbsInfer::Project::RubyRuntimeGenerator.new(app_dir: Rails.root.to_s).generate
       puts "Generated language-runtime pseudo-code: #{dir}"
     end
   end
--- a/lib/rbs_infer/tasks/rbs_infer_ruby_runtime.rake
+++ b/lib/rbs_infer/tasks/rbs_infer_ruby_runtime.rake
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require_relative "../project/ruby_runtime_generator"
+
+namespace :rbs_infer do
+  namespace :ruby_runtime do
+    desc "Generate the language-runtime pseudo-code sidecar for Steep (sig/generated/steep_ruby_runtime/)"
+    task :all do
+      app_dir = defined?(Rails) ? Rails.root.to_s : Dir.pwd
+      dir = RbsInfer::Project::RubyRuntimeGenerator.new(app_dir: app_dir).generate
+      puts "Generated language-runtime pseudo-code: #{dir}"
+    end
+  end
+end
--- a/spec/bin/rbs_infer_spec.rb
+++ b/spec/bin/rbs_infer_spec.rb
@@ -1,4 +1,5 @@
 require "spec_helper"
+require "rbs_infer/project/ruby_runtime_generator"
 require "tmpdir"
 require "fileutils"
 require "open3"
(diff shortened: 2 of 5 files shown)


lib/rbs_infer/tasks/rbs_infer_ruby_runtime.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require_relative "../project/ruby_runtime_generator"
+
+namespace :rbs_infer do
+  namespace :ruby_runtime do
+    desc "Generate the language-runtime pseudo-code sidecar for Steep (sig/generated/steep_ruby_runtime/)"
+    task :all do
+      app_dir = defined?(Rails) ? Rails.root.to_s : Dir.pwd
```
felixefelip: A constante Rails não está sempre definida aqui?
```suggestion
      app_dir = Rails.root.to_s
```

spec/bin/rbs_infer_spec.rb
```diff
@@ -453,6 +454,10 @@ def setup_concern_project
         end
       RUBY
 
+      # A transcrição da linguagem, que todo projeto tem em disco: sem ela o
```
felixefelip: Remova esse comentário

lib/rbs_infer/tasks/rbs_infer_ruby_runtime.rake
felixefelip: Os dois aplicados em `e79e7d9`.

**Sobre o `Rails`:** voce esta certo, e fui verificar antes de aceitar. Nada alem do Railtie carrega esses `.rake`, e `rake_tasks do` so roda dentro de app Rails - entao `defined?(Rails)` e sempre verdadeiro e o `: Dir.pwd` era braco morto.

```ruby
dir = RbsInfer::Project::RubyRuntimeGenerator.new(app_dir: Rails.root.to_s).generate
```

Eu tinha copiado o guarda dos vizinhos sem perguntar se ele fazia algo. Nao faz - e **os outros nove `.rake` carregam o mesmo**. Nao mexi neles para nao alargar o diff da revisao; e um `sed` num PR de limpeza, se quiser.

Uma consequencia que vale registrar: isso fecha a porta para um projeto **sem Rails** carregar esse arquivo a mao, o que e concebivel porque o `RubyRuntimeGenerator` e core. Hoje ninguem faz, e projetar para um caso que nada exercita foi justamente o que produziu o braco morto. Se um dia existir, o `Makefile` ja mostra o outro caminho: chamar o gerador direto.

**Comentario removido** como pedido.

Verificado: `rake rbs_infer:ruby_runtime:all` continua gerando o sidecar, e a suite segue em 1579 exemplos, 0 falhas.

Issue #321, opened 2026-09-07
Carry a stored block's body in its type (RbsInfer::Body), so a replay reads a value instead of walking shapes


felixefelip:
## What this proposes

Give a block literal a type that carries **its own source**, so `class_eval(&blk)` is resolved by reading a value's type instead of by recognising the DSL chain that carried the block there.

```rbs
module RbsInfer
  class Body[B] < ::Proc
  end
end

class Example34
  module Foo
    @_bazingado_block: (^(*untyped) [self: singleton(Example34::Bar) | singleton(Example34::Baz) | Module] -> Symbol
                        & ::RbsInfer::Body["\n      def age\n        31\n      end"])?
  end
end
```

The wide `^(*untyped) [self: ...] -> Symbol` is what every `Proc` consumer already understands; `& Body[...]` is a payload only rbs_infer reads. A checker that has never heard of `Body` sees a `Proc`.

## Where it comes from

`StoredBlockReplayExpander` answers exactly one question — *which block reaches this `class_eval`* — and answers it with a hand-rolled interprocedural dataflow. `ShapeReader` (four `*_replay_shape` readers), `CallGraph`, `DeferralReader`, `Resolution`, `Shapes`, `ShapeSet` and `ReaderCollector` are ~1600 of the module's 2815 lines, and all of them exist to walk from a call site back to a block.

That is the shape `docs/engineering/prefer-principled-over-shortcuts.md` describes: machinery that re-derives, syntactically, what the pipeline already does semantically. If the block value carried its own identity, the walk is a lookup.

## Why raw source, and not the members it defines

The structural alternative is for the type to say what the block *installs*:

```rbs
^() -> void { def age: () -> Integer }
```

It does not close. Writing `def age: () -> Integer` requires having already inferred the block's body **in the scope it is lexically written in**, with the wrong `self` — which is precisely what the relocation exists to fix. Raw source defers inference to the replay site, where target, definee and self are all known, and it is lossless for the things return inference needs: method bodies, ivar writes, calls to the host's own methods, nested blocks.

This does **not** reintroduce felixefelip/rbs_infer#156. The payload is a purely syntactic slice of the source — the same slice `BlockReopen.appended` already takes — so nothing here reads rbs_infer's own generated output before the parse.

## Why a copy, and not a reference

`Body[::Example34::Baz, :included]` — an identity that re-reads the live source, the same `(in, call)` key `StoredBlockReplayImplements` already uses — was the alternative. Rejected:

[...]

PR #322, opened 2026-09-07
Carry a stored block's own source in its slot's type (RbsInfer::Body)

felixefelip:
Closes the first step of #321.

A slot that holds a later-replayed block now declares what it holds:

```rbs
@_bazingado_block: (^(*untyped) [self: singleton(Example43::Bar) | …] -> Symbol
                    & ::RbsInfer::Body["def age\n        31\n      end"])?
```

The wide `^(*untyped) [self: …] -> Symbol` is exactly what inference already wrote — every `Proc` consumer understands it, and a checker that has never heard of `Body` sees a `Proc`. The intersection is a payload only rbs_infer reads.

## How the block is found

Off the **same replays the rewrite is built from**, never re-derived. `Storage` names `(owner, method, ivar)` and a `Replay`'s `call` is that same storage method, so the method is the join — and a block the resolution declined does not reach a slot here either, so the payload and the relocation cannot disagree about which block is which. Bodies are sliced against `replay.source` rather than the file being expanded, for the reason `Replay#source` exists at all: a concern's block is written in the concern's file while the `include` that resolves it is written in the host's.

`Collector` is untouched apart from one public reader for `storages`; `shapes` stays protected.

## What this does NOT do

**It carries the block; it does not attribute it.** `Example43` is where the two come apart:

```ruby
class Bar;       bazinga(Example43::Baz)                        # receives: age
class BarOther;  bazinga(Example43::Baz)                        # receives: age
class BarOther2; bazinga(Example43::Baz, Example43::BazOther)   # receives: age + name
```

The ivar is **one line declared in `Foo`**, so its type can only say `Body[age] | Body[name]` — and that union is wrong at each end: it says "one of these" to a class that got exactly one, and "one of these" to a class that got both, which a union of two cannot express at all.

So nothing may read `Body[…]` off this declaration and conclude which host received which block. That answer stays with `Resolution`, per call site, and RBS has no declaration site for a call. `decorate_stored_block_ivars` carries that caveat in full so the next reader does not mistake the payload for an attribution.

This is the measurement #321 asked for, and it is worth stating plainly: **the ivar is the wrong attachment point** for the larger goal of replacing the shape walk. What lands here is the block's source travelling with its slot.

## `::RbsInfer::Body`

Declared by the CLI into the output dir (`sig/rbs_infer/rbs_infer_body.rbs`), in the same shape as the postconditions sidecar — written when something carries a payload, removed when nothing does. Without it `steep check` fails with `Cannot find type ::RbsInfer::Body` on every file that carries one.

## Escaping

Only `\`, `"`, `\n` and `\t`; `#` is deliberately left alone. RBS string literals do not interpolate, and `String#inspect` — which escapes `#` before `{` — walks into an rbs unescape bug that returns the literal with a NUL injected (`"a \#{now} b"` comes back as `"a \#{now} b"`, measured on rbs 4.2.1.pre.1). Round-trip verified for escaped quotes, double backslashes, newlines, tabs, `$globals`, `@ivars` and unicode.

## Verification

- **1581 examples, 0 failures**, including the dummy's `steep check` against its baseline.
- The emitted RBS parses and both bodies come back out of the type intact, `"John Doe"` included.
- The dummy's `sig/` was taken to a fixed point (`make steep` → `make rbs_infer` until two passes hash identically) **before** the expectations were regenerated, since the suite reads that `sig/` as input. Getting that order wrong is what made the failing set move between runs.

## For review

In `example32/33/39/42` the DSL method's return also loses its `Symbol` member (`-> (^… | nil | Symbol)` becomes `-> (^…)?`). **That is not from this change** — isolated with the decoration disabled, the same output appears. It is pre-existing drift between those checked-in files and HEAD, updated here in passing.

Not included, deferred from #321: a dedicated unit spec for the escaper. It is exercised only through the generated output today.


lib/rbs_infer/analyzer.rb
```diff
@@ -314,11 +320,69 @@ def build_single_target_rbs
...
+  # union of both — while the relocation correctly gives `Bar` just `age` and
+  # `BarOther2` both. The union is wrong at each end: it says "one of these" to
+  # a class that got exactly one, and "one of these" to a class that got both.
+  #
+  # So nothing may read `Body[…]` off this declaration and conclude which block
+  # a given host received — that answer comes from `Resolution`, per call site,
+  # and RBS has no declaration site for a call. What this emission is good for
+  # is the block's SOURCE travelling with the slot; the attribution stays where
+  # it is (felixefelip/rbs_infer#321).
+  def decorate_stored_block_ivars(ivar_types, module_ivar_types)
```
felixefelip: esses métodos novos não poderiam ficar numa outra classe para manter o analyzer com poucas linhas?

lib/rbs_infer/project/stored_block_replay_expander.rb
```diff
@@ -122,6 +122,56 @@ def expand(source, sources:, mixin_index:)
       apply_replays(source, replays, extensions, mixin_index)
     end
 
+    # The block literals this file's storage slots hold, as
```
felixefelip: Remova esses comentários

lib/rbs_infer/project/stored_block_replay_expander/collector.rb
```diff
@@ -27,6 +27,12 @@ class Collector < Prism::Visitor
     # site here does to a class here, read off the same resolution.
     attr_reader :extensions
 
+    # The slots this file's methods keep a block in. Public where `shapes` is
```
felixefelip: remova esses comentários

lib/rbs_infer/analyzer.rb
felixefelip: Movido para `RbsInfer::Project::StoredBlockIvarDecorator` (novo arquivo em `lib/rbs_infer/project/`). Saíram 56 linhas do Analyzer — os quatro métodos e o comentário longo sobre a ressalva do Example43, que agora mora na classe. O Analyzer ficou com uma chamada.

Fiz classe e não módulo seguindo a regra que o `DeferralReader` já documenta neste pass: ela segura uma dependência (o mapa de corpos), que assim não aparece na assinatura de nenhum dos métodos. O `BlockBodyType` continua módulo porque não segura nada.

lib/rbs_infer/project/stored_block_replay_expander.rb
felixefelip: Removidos. Você ancorou na linha 125 (o bloco do `stored_block_bodies`); tirei também o do `bodies_by_slot` logo abaixo, por ser a mesma adição. Se era só o primeiro, eu recoloco o segundo.

lib/rbs_infer/project/stored_block_replay_expander/collector.rb
felixefelip: Removido.

PR #325, opened 2026-09-07
Root the `extend` an inward hook writes back into the host

felixefelip:
Found by running `rbs_infer` on a real app after #324: several controllers lost their ivar declarations outright — `@export`, `@board`, `@card`, `@join_code`, `@users` — and `BoardsController#@board` lost a union member. Not widened to `untyped`, *gone*.

## What broke

`StoredBlockReplayExpander` says what a hook's `base.extend(M)` did by reopening the host with that one line:

```ruby
class Account
  extend Storage::Totaled::ClassMethods
end
```

The name it writes is an **answer**. `Declarations#resolve` already decided, absolutely, which declaration the hook hands over. But the reopening it lands in is a **lexical scope**, so a relative name gets resolved a second time — this time under the host. `Account` has an `Account::Storage` (the concern that pulls `Storage::Totaled` in), so the line named `Account::Storage::Totaled::ClassMethods`, which nothing declares. Ruby itself would resolve it the same way and raise.

What that costs is not the mixin but the class. RBS cannot build a singleton with a mixin it cannot find:

```
RBS::NoMixinFoundError: sig/generated/app/models/account.rbs:2:2...2:39:
  Could not find mixin: Storage::Totaled::ClassMethods
```

so `Account`'s singleton failed wholesale, `Current.account` came out `untyped`, and every `@x = Current.account.…` write had no type for `IvarWriteAnalyzer` to record. The ivar's type set then held only `nil`, which `infer_ivar_types` reads as "Steep saw nothing nominal" and drops — hence a *missing* declaration rather than a wide one. The sibling `extend ::Account::MultiTenantable::ClassMethods` was fine only because its first segment happens to be `Account`.

Before / after on the app's `Account::ExportsController`:

```diff
 class Account
   class ExportsController < ApplicationController
-    CURRENT_EXPORT_LIMIT: Integer
+    @export: (Account::Export | (Account::Export & Account::Export::Validated))?
+
+    CURRENT_EXPORT_LIMIT: Integer
```

## The fix

The line goes in rooted: `extend ::#{extension.name}`. `Declarations#resolve` never returns a `::`-prefixed name, and `running_selves` already writes `::#{host}` for the same reason — the name resolved is the name written.

## What pins it

`example64` — a host with a nested module named `Vault`, including a top-level `::Vault::Totaled` concern whose `class_methods do` block defines `vault_key`. `Example64#key` is `String` only when the host's singleton carries the concern's class methods; with the fix reverted it regenerates as `untyped`, and `steep check` on the dummy reports `Type singleton(::Object) does not have method vault_key`.

A unit spec (`roots the name against a target that shadows its namespace`) states the same thing at the expander's own level, and fails without the fix.

The nine other expanded snapshots move by exactly one `::` each.

## Verified

- `bundle exec rspec` — 1582 examples, 0 failures.
- Steep baseline and `.steep_contracts.yml` unchanged: the new fixture adds no diagnostics.
- Re-ran the app: `Account`'s RBS builds, `@export` is back byte-identical to what it was before the regression.


lib/rbs_infer/project/stored_block_replay_expander.rb
```diff
@@ -257,8 +257,20 @@ def union(parts)
     # no block to slice and nothing to re-indent, only the one line the hook
     # runs — but it goes in the same place, for the same reason, and is dropped
     # on a second pass by the same `missing_from`.
+    #
```
felixefelip: Remova esse comentário

PR #328, opened 2026-09-08
Write the ActiveJob enqueue round trip as pseudo-code

felixefelip:
A job's `perform` is never called by the app; `MyJob.perform_later(user)` is. The gem RBS types that as `(*untyped) -> untyped`, so every `perform` parameter infers `untyped` however unambiguous the call site — `spec/dummy`'s only job comes out `def perform: () -> untyped`.

What ActiveJob does between the two is a plain-Ruby forward, so write it as one. `RbsInfer::Extensions::Rails::Jobs::RuntimeGenerator` emits a single reopen of `ActiveJob::Base` under `sig/generated/steep_activejob_runtime/`, gated on the app actually subclassing it:

```ruby
class ActiveJob::Base
  def self.perform_later(*args, **kwargs)
    new.perform(*args, **kwargs)
    new
  end
end
```

Nothing there states a type. The trailing `new` is the enqueued job `perform_later` really returns, not `perform`'s value — without it the enqueue call's type leaks into every caller's return.

## Known limitation, deliberate

This does not type any `perform` yet. Written once on the base class, the reopen is shared by every job in the app, so the call-site evidence lands on **that** class's `*args`. Measured on a dummy fixture (a `perform(user, message:)` job with one call site):

```rbs
class ActiveJob::Base
  def self.perform_later: (*(User & User::Validated | { message: String }) args, **untyped) -> ActiveJob::Base
end

class NotifyUserJob < ApplicationJob
  def perform: (untyped user, message: untyped) -> untyped   # unchanged
end
```

Two jobs would feed that same parameter, and `new.perform` in this body has no way back to the job that was called. Closing it needs the call site's receiver to reach the forward: either instantiating this body per subclass, or teaching `InvokerSelfTypes` to resolve a call made *on* a receiver (it declines those today).

For the record, the same forward written at the call site does work end to end, with no core change:

```ruby
NotifyUserJob.perform_later(u, message: "hi")  →  NotifyUserJob.new.tap { |job| job.perform(u, message: "hi") }
# def perform: (User & User::Validated user, message: String) -> String
```

That shape is not what this PR ships; it is noted as the reference for whichever follow-up closes the gap.

## Wiring

`rake rbs_infer:job_runtime:all` (railtie), `make rbs_infer_job_runtime`, and a row in `make rbs_generators_all` — same molde as the controller/view/current runtimes. README documents the generator and the limitation.

## Testing

- 7 new unit examples for the generator (shape, gating, stale-dir removal, output path).
- Full suite: 1589 examples, 0 failures.
- `make steep` on the dummy adds exactly one diagnostic — `ActiveJob::Base#perform` is declared `() -> untyped`, so the forwarded arguments are unexpected — absorbed into `spec/expectations/steep_baseline.txt` as the record of the gap. The PR that closes it is the one that removes it.


lib/rbs_infer/extensions/rails/jobs/runtime_generator.rb
```diff
@@ -0,0 +1,128 @@
+# frozen_string_literal: true
+
+require "fileutils"
+require "prism"
+
+module RbsInfer
+  module Extensions
+    module Rails
+      module Jobs
```
felixefelip: Por que não seguir os nomes originais do Rails?
```suggestion
      module ActiveJob
```

spec/dummy/sig/generated/steep_activejob_runtime/active_job_base.rb
```diff
@@ -0,0 +1,16 @@
...
+#
+# What ActiveJob does between `perform_later` and `perform`, as plain
+# Ruby: the arguments handed to the enqueue call are the arguments
+# `perform` receives, on a fresh instance. The trailing `new` is the
+# enqueued job `perform_later` returns.
+
+class ActiveJob::Base
+  def self.perform_later(*args, **kwargs)
+    new.perform(*args, **kwargs)
+    new
```
felixefelip: No código original do activejob ele faz esses 2 news?

lib/rbs_infer/extensions/rails/jobs/runtime_generator.rb
felixefelip: Nenhuma razão boa — era inconsistência minha. O vizinho já é `Rails::ActiveRecord::RuntimeGenerator` em `active_record/`, então adotei o mesmo nome: `Rails::ActiveJob::RuntimeGenerator` em `active_job/`. Task, Makefile, README e spec acompanham (04a8093).

spec/dummy/sig/generated/steep_activejob_runtime/active_job_base.rb
felixefelip: Não faz — bem visto. O `perform_later` real instancia uma vez só e devolve essa mesma instância:

```ruby
def perform_later(...)
  job = job_or_instantiate(...)
  enqueue_result = job.enqueue
  yield job if block_given?
  enqueue_result
end
```

O segundo `new` devolvia um job que nunca tinha recebido os argumentos. Agora é `job = new; job.perform(*args, **kwargs); job` — as mesmas três linhas, com a fila entre as duas pontas colapsada (04a8093).

PR #333, opened 2026-09-08
Follow a dispatcher inherited from an ancestor, keeping the receiver

### LINKED ISSUE #331: Infer a method's parameters through a forward inherited from an ancestor
## The shape

A method a subclass implements, invoked only through a dispatcher the base class defines. ActiveJob is the case at hand — `perform` is never called by the app, `MyJob.perform_later(user)` is — but the pattern is the plain template method:

```ruby
class Base
  def self.dispatch(*args, **kwargs)
    handler = new
    handler.handle(*args, **kwargs)
    handler
  end
end

class Concrete < Base
  def handle(user, count:); end   # every parameter infers `untyped`
end
```

Every argument is unambiguous at the call site. A human reads `Concrete.dispatch(user, count: 3)` and knows what `handle` receives. The pipeline does not.

## Where the evidence is lost

The call site *is* accepted, by ancestry, in `NewCallCollector#ancestry_match_key` (`lib/rbs_infer/inference/new_call_collector.rb:596`):

```ruby
def ancestry_match_key(component, method_name)   # component = "AuthorDigestJob"
  owner = method_owner_on_either_side(component, method_name)   # => "ActiveJob::Base"
  return method_name if normalized_owner == @target_class       # <- receiver discarded here
```

The concrete receiver is in hand and dropped, because the key is the bare method name. `ParamTypeInferrer#infer_from_callers` then merges every usage under that key, so all subclasses cross into one parameter. `spec/dummy` pins the result (#329) — two jobs enqueued from two controllers put four unrelated branches on the shared forward:

```rbs
def self.perform_later: (*(Post | { recipient: String? } | User & User::Validated | { sizes: Array[Integer] }) args, **untyped) -> ActiveJob::Base
```

while both `perform` methods stay `untyped`. No `perform` could accept that union, which is the defect stated in RBS.

## That a per-subclass forward closes it, measured

Writing the forward by hand for ONE of the two dummy jobs, with the other as the control in the same run:

```rbs
# with its own forward            # inheriting the base one (control)
AuthorDigestJob#perform:          AvatarThumbnailJob#perform:
  (Post post, recipient: String)    (untyped user, sizes: untyped)
```

No core change was needed for that, because a same-class forward matches by NAME (`match_class_branch?`) instead of by ancestry. So the missing capability is precisely: *follow a forward that lives in an ancestor, keeping the call site's receiver*.

Generating one forward per job would work and is the fallback, but it abandons the single pseudo-code sidecar (#328) and scales with the number of jobs.

## Proposal

Generalize the wrapper-forwarding inference that already exists — `ParamTypeInferrer#infer_wrapper_method_param_types` / `#detect_forwarding_methods` (#46), which today types a method from the call sites of the wrapper forwarding into it. Three limits to lift:

1. **The wrapper may live in an ANCESTOR**, not only in the target's own parse. The ancestor's source is reachable through `SourceIndex` like any other class.
2. **The forward may be `new.<method>(...)` with positional arguments**, not only `Klass.new(kw: param)` — `detect_forwarding_methods` currently requires a `CallNode` named `:new` *with a receiver* and reads only `KeywordHashNode` elements, so a receiverless `new` forwarding positionals matches nothing.
3. **The wrapper's call sites are filtered by RECEIVER.** This is the essential piece and the one that makes the result correct rather than merely non-empty: only `Concrete.dispatch(...)` sites feed `Concrete#handle`. `receiver_components` / `match_class_branch?` already do this matching.

The flow stays as it is today — the target class pulls its own evidence — so nothing has to write facts about another class.

## Why this over a per-subclass generator

It is not a rule about ActiveJob. It closes the template-method-on-a-base-class pattern generally, and it keeps the sidecar a single reopen of `ActiveJob::Base`.

## Acceptance

- `AuthorDigestJob#perform` infers `(Post post, recipient: String)` and `AvatarThumbnailJob#perform` infers `(User & User::Validated user, sizes: Array[Integer])` — the two snapshots added in #329, which exist to be updated by this change.
- The two jobs do NOT cross: each takes only its own enqueue site's types. The two-job fixture is what detects a receiver filter that is too loose.
- The sidecar stays one reopen of `ActiveJob::Base` (#328).
- No new steep diagnostics; the baseline entry for the sidecar's forwarded arguments should become removable.

## Risk

This touches the core of parameter inference, and a receiver filter that is too loose silently re-crosses unrelated callers instead of failing — the failure mode the fixture in #329 was built to catch.

felixefelip:
Closes #331.

A method a subclass implements, invoked only through a dispatcher the base class defines, had every parameter infer `untyped` — however unambiguous the call site. ActiveJob is the case that surfaced it (`perform` is never called by the app, `MyJob.perform_later(user)` is); `example65` is the same shape with no framework in it.

## The one line where it was lost

The call site was already *accepted*, by ancestry, in `NewCallCollector#ancestry_match_key`:

```ruby
owner = method_owner_on_either_side(component, method_name)  # => "ActiveJob::Base"
return method_name if normalized_owner == @target_class      # ← receiver dropped
```

The concrete receiver is in hand there and discarded, because the key is the bare method name. Every subclass's arguments then merged into the base's `*args` under that one key, and no handler got any.

## What was added

`InheritedForwards` answers the missing question: which dispatchers does this class **inherit**, and which of its own methods does each hand its arguments to. `NewCallCollector` then reads `Greeter.dispatch("ada")` as a call site of `Greeter#handle`, mapping the arguments exactly as a direct `greeter.handle("ada")` is mapped.

Two rules keep it honest:

- **The forward must be transparent** — its arguments are the rest and keyrest, splatted, and nothing else. An argument added, dropped, read instead of splatted, or taken from another local breaks the position-for-position correspondence, and then this answers nothing rather than guessing. Nine unit examples pin that rule.
- **The receiver must reach the target's handler** — `method_owner(component, forwarded_to) == @target_class`, which is the same question `ancestry_match_key` asks of a direct call, so the two paths agree. Not "the receiver IS the target": a subclass that adds nothing still runs the target's handler, and discarding its dispatch site leaves a type that is too narrow rather than absent. The receiver must also be spelled as a class (constant, constant path, `self`, or a `singleton(...)` type), since a dispatcher is inherited onto the class — otherwise an instance method sharing the forward's name would deposit its arguments on the handler.

Recognized spellings are `def self.x` and `class << self`. A module's **instance** method — the only spelling `extend` actually carries to a host — and `class_methods do` (which reaches the pipeline expanded, while this reads the raw source `ParseCache` holds) are **not** covered, and are noted as open in the source.

## Result

Four parameters that were `untyped`:

```rbs
Example65Greeter#handle:    (String name, greeting: String) -> String
Example65Adder#handle:      (Integer count, step: Integer) -> Integer
AuthorDigestJob#perform:    (Post post, recipient: String) -> ...
AvatarThumbnailJob#perform: (User & User::Validated user, sizes: Array[Integer]) -> ...
```

Each pair takes unrelated types on purpose (#329, #332): each handler takes only its **own** call site's arguments, which is exactly what a receiver filter too loose would break. Regenerating the whole dummy moves those four files and nothing else.

The base's own `*args` still shows the union of every dispatch site — it genuinely receives all of them — and the ActiveJob sidecar stays one reopen of `ActiveJob::Base`, which was the point of writing it there in #328.

## Threading

`inherited_forwards` is **required**, not defaulted: a forgotten wire gets the old behaviour, which reads as an answer rather than failing. It caught two missing call sites while I was making this change (`docs/engineering/required-threaded-deps.md` earning its keep). The two `MethodTypeResolver` walks and `find_new_calls` pass `{}` with a stated reason — they keep `.new` usages and discard method calls, so looking for a forward there would only cost a sweep.

## Cost

Measured rather than argued: **941 sweeps per full dummy run, 51ms total — 0.13% of a 38s run**. Left unmemoized on those numbers; per-sweep cost is ~54µs, so it is file count that would move it.

## Testing

- 17 unit examples on the transparency rule: the dispatcher keeping a parameter of its own (required / optional / named keyword), a block parameter being fine, the forwarded values touched before the call (reassigned, `shift`ed, `delete`d), the same arguments reaching two handlers, `class << self`, a local not always holding a fresh instance, another class's instance, no rest to forward, instance method, wrong method name, several forwards — plus `for_methods` keeping every method a forward reaches.
- Three fixtures for what nothing covered: `Example65Reporter` (typed **only** through its empty subclass `Example65CsvReporter`), and `Example65Impostor` — a second hierarchy spelling its dispatcher identically, whose `bool`/`Symbol` arguments would surface in the Greeter's snapshot if the receiver ever stopped deciding which class is reached.
- Two spec builders got a test-only `{}` default; the production API stays strict.
- Full suite: 1621 examples, 0 failures. Steep baseline unchanged.

Review feedback on the first cut is addressed in ae3d6fb — see the [reply](https://github.com/felixefelip/rbs_infer/pull/333#issuecomment-5591803623).


felixefelip:
## Review

The core idea is right — the defect was `ancestry_match_key` dropping the receiver, and keeping it is the fix. Two things need work before this lands: **the transparency rule is only checked on the call node**, and the receiver filter traded "too loose" for "too strict".

Everything below was confirmed by running `InheritedForwards.transparent_forwards` against the source shown, not read by eye.

---

### 1. Transparency ignores what happens *before* the forward → wrong types

`transparent_forward?` requires the *call* to splat `rest`/`keyrest` and nothing else. It never requires the `def` to take **only** rest and keyrest, and never checks the local was not reassigned or mutated in between. Probe output:

```
leading required + rest         => ["dispatch"]   # accepted
args = args.drop(1) before it   => ["dispatch"]   # accepted
args.shift before it            => ["dispatch"]   # accepted
kwargs.delete before it         => ["dispatch"]   # accepted
```

The simplest case needs no mutation at all:

```ruby
class Dispatcher
  def self.dispatch(tag, *args, **kwargs) = new.handle(*args, **kwargs)
end

Greeter.dispatch(:audit, "ada", greeting: "hi")   # handle receives "ada"
```

`extract_cross_class_args` maps the `dispatch` node positionally against `handle`'s parameters, so `handle`'s `name` gets `Symbol`. That is not imprecision, it is the wrong answer — exactly what `extract_cross_class_args` already argues about splats in its own comment ("a wrong parameter type is worse than none, because it reads as an answer"). `args.shift` / `args = args.drop(1)` are how a real dispatcher peels off its own control argument, and they land in the same place.

The rule has to be about the whole `def`, not the call: require `requireds`/`optionals`/`keywords` to be empty (rest + keyrest, plus `&blk`), and decline when `rest`/`keyrest` is written to, or used anywhere other than as the forward's splat (`args.shift` is a `CallNode` whose receiver is `args` — detectable).

### 2. The new branch does not require a singleton receiver

```ruby
if !@inherited_forwards.empty? && node.receiver && node.arguments
  forwarded_to = @inherited_forwards[node.name.to_s]
```

`dispatch` is a class method, but the match is name + `match_class?`, and `match_class?` is just as happy with a local typed `Greeter`. If the class (or an ancestor) has an **instance** method sharing the forward's name — plausible for `run`, `call`, `process`, `dispatch` — then `x.run(cfg)`, an unrelated and entirely legitimate call, deposits `cfg`'s type on the handler's first parameter. Worth requiring `node.receiver` to be a `ConstantReadNode`/`ConstantPathNode`/`SelfNode`, or the resolved type to be spelled `singleton(...)`.

### 3. `match_class?` drops a descendant receiver — and the ordinary path does not

`match_class?` matches only an identical name or a relative suffix. So:

```ruby
class BaseImportJob < ApplicationJob
  def perform(path); end
end
class CsvImportJob < BaseImportJob; end

CsvImportJob.perform_later("/tmp/x.csv")   # runs BaseImportJob#perform
```

The only call site of `BaseImportJob#perform` is discarded → `(untyped path)`. Note the asymmetry: for a direct `x.perform(path)` with `x : CsvImportJob`, `ancestry_match_key` **accepts** it (`new_call_collector.rb:632` — owner == target returns `method_name`). The new branch is stricter than the path it complements, and under the whole-program assumption a missed call site yields a type that is too narrow, not merely absent.

The right predicate is not "the receiver IS the target" but "the `handle` this receiver reaches belongs to the target": `rbs_definition_resolver.method_owner(component, forwarded_to) == @target_class`. That covers the descendant and still excludes the base (`Dispatcher` has no `handle` of its own here — and where it does, the owner is itself, not the target).

### 4. Class-method spellings the syntactic filter cannot see

`transparent_forwards` requires a `Prism::DefNode` with `receiver.is_a?(SelfNode)`:

```
class << self                       => []
concern class_methods block         => []
extend-able module (def dispatch)   => []
anonymous splat forwarding (*, **)  => []
... forwarding                      => []
```

Two things matter more than the list itself:

- **The `extend` claim in the description cannot fire.** A `def self.dispatch` inside a `module` lives on the *module's* singleton, and `extend` does not copy it to the host. The only spelling that reaches a host through `extend` is a module **instance** method (`def dispatch`) — which is precisely the one the `SelfNode` filter rejects. The `inherited?` half would cover it (`method_owner` returns the module, and `declared_classes` contains it); it is the syntactic half that blocks it.
- `class << self` and `class_methods do` are the dominant Rails spellings for this exact pattern.

Related: `ParseCache` reads **raw** source (`SourceReader.read`, no `SourceExpanders`), while expanders only run over the target in `Analyzer`. So a dispatcher defined through `class_eval` — which the project supports via `SelfClassEvalExpander` — is invisible here even though RBS reports the owner correctly. The two halves disagree about what the program is.

### 5. One forward into two handlers: last one wins, silently

`for_methods` does `acc[forward] = method_name`. With the full template method:

```ruby
def self.run(*args, **kwargs)
  h = new
  h.setup(*args, **kwargs)
  h.handle(*args, **kwargs)
  h
end
```

`transparent_forwards` returns `["run"]` for both `"setup"` and `"handle"` (confirmed), and the Hash keeps only one — which one depends on `target_methods.keys` order. The other handler stays `untyped` with no signal. The value should be a list.

### 6. ActiveJob: the two most common enqueue forms do not reach it

- `MyJob.set(wait: 5.minutes).perform_later(user)` — the receiver is a `CallNode`, `resolve_method_chain` gives `ActiveJob::ConfiguredJob` (or nothing), `match_class?` fails. Probably the most common form in a real app.
- `MyJob.perform_now(user)` — the #328 sidecar only writes `perform_later`. Not this PR's defect, but this PR is what makes the gap visible, and a `def self.perform_now(*args, **kwargs) = new.perform(*args, **kwargs)` in the sidecar would be picked up by the new machinery for free.

### 7. Cost: a full AST sweep per (target, method, calling file), unmemoized

`forwards_into` runs `find_all_nodes` over the root of **every** file `files_calling(name)` returns, for **every** parameterized method of **every** target — and the sweep happens *before* `inherited?`, so it is paid in full even when the target inherits nothing. Measured on the dummy: **905 full-file sweeps per run** (210 parameterized defs). In a real app the hot names are `save`, `call`, `update`, `perform`; `files_calling`'s own comment warns it is a candidate filter, not a match. Nothing is memoized across targets even though `(file, method_name)` always has the same answer. A `[[file, method_name]]` memo — or a single per-file index of "singleton defs with a rest/keyrest" — removes nearly all of it. Given `docs/tasks/performance_optimizations.md`, worth measuring before merge.

### 8. Smaller

- **`inherited?` has no test at all.** The nine new examples cover only the syntactic half. The description says the RBS half is "covered by the example65 integration snapshots", but example65 has no **negative** case — no unrelated class with a same-named `def self.dispatch` forwarding into `handle`. Deleting the `declared_classes` check today breaks nothing. That is the test that actually protects the fix.
- **`new_locals` is not flow-sensitive:** `handler = new; handler = Other.build; handler.handle(*args)` is accepted (confirmed). Contrived, but the local stays marked forever.
- **`splats_exactly?`** uses class objects as tags and then does `case kind.name when "Prism::SplatNode"` — a string comparison on the name of a class already in hand. Two plain `if`s say the same with less indirection. Also `EMPTY` is frozen but `for_methods` returns a fresh unfrozen `{}` on the normal path.
- **Bootstrap:** `inherited?` depends on the target's own RBS existing (`method_owner("singleton(Target)", …)`). On a cold single-file run with no prior `sig/`, the feature simply does not fire — consistent with the stabilization loop, but worth a line so it does not read as "works sometimes".

---

**Summary:** 1, 2 and 3 look blocking — 1 and 2 emit wrong types from valid code, 3 emits a too-narrow type for an everyday Rails shape (base job + empty subclass) and is a precision regression against the path that already existed. 4 and 5 are silent false negatives, and 4 invalidates a claim in the description. 6 and 7 are follow-up, though 7 is worth measuring first.

https://claude.ai/code/session_01NDmjVN7kzLv7bjc7wdQ815

felixefelip:
Every claim reproduced before changing anything — 1, 2, 4, 5 and the `new_locals` note with probes against `transparent_forwards`, 3 by reading `ancestry_match_key:632` against my own filter. All of it stands. Fixed in ae3d6fb.

**1 — transparency was a claim about the call node.** Confirmed, and worse than imprecise: `dispatch(tag, *args, **kwargs)` typed `handle`'s first parameter `Symbol` from the caller's control argument. The rule is now about the whole `def` — no parameters besides rest, keyrest and a block; no write to what is forwarded; and every read of it must be a **splat**, which hands over the elements and never the collection, so no callee can reach back into it. Picking "splat" rather than "inside this call node" matters for your own case 5: `h.setup(*args); h.handle(*args)` has to stay legal, and it does.

**2 — no singleton requirement.** Confirmed. A dispatcher is inherited onto the class, so the receiver must be spelled as one (constant, constant path, `self`, or a type spelled `singleton(...)`).

**3 — the descendant receiver, and the asymmetry.** This was the one I would have missed; the asymmetry argument is what makes it obviously wrong rather than a judgement call. The predicate is now yours: `method_owner(component, forwarded_to) == @target_class`, asked of the same resolver `ancestry_match_key` uses, so the two paths agree. `Example65Reporter` + an empty `Example65CsvReporter` is now a fixture, and it was `(untyped path)` before.

**4 — spellings, and the `extend` claim.** `class << self` is in. You are right that the description's `extend` claim could not fire: a module's `def self.dispatch` lives on the module's own singleton, and the only spelling `extend` carries to a host is the module **instance** method my `SelfNode` filter rejects. I have corrected the description rather than leaving it. That one and `class_methods do` (which needs the expanded source, not the raw `ParseCache` — your `class_eval` point is the same disagreement about what the program is) are noted in the source as open, not claimed.

**5 — last one wins.** Confirmed: `transparent_forwards` returns `["run"]` for both `setup` and `handle`, and the Hash kept one. The value is a list now.

**7 — measured.** 941 sweeps on a full dummy run (your 905, plus the new fixtures), **51ms total — 0.13% of a 38s run**. So I have left it unmemoized and recorded the numbers rather than optimizing on the shape of the code. Your point about hot names in a real app stands and is the right trigger to revisit; the per-sweep cost is ~54µs, so it is file count that would move it.

**8 —** `new_locals` is flow-sensitive now (every assignment to the name must be a `new`). `splats_exactly?` no longer tags with class objects and then compares their names. The bootstrap is stated in the class comment. On the missing negative test: rather than a unit test of `inherited?` alone, `Example65Impostor` is a second hierarchy spelling its dispatcher identically — its `bool`/`Symbol` arguments would surface in the Greeter's snapshot the moment any filter stopped proving which class the receiver reaches. That is the test that fails loudly, and it covers the collector's half too.

**6 —** left for follow-up, as you had it. `perform_now` in the #328 sidecar is a one-line change that this machinery would pick up for free; `.set(wait:).perform_later` needs the receiver to resolve through `ConfiguredJob` and is a real piece of work. I will open an issue rather than grow this PR.

Suite: 1621 examples, 0 failures (17 on `InheritedForwards`, up from 9). Steep baseline unchanged.

PR #343, opened 2026-09-10
feat(core): render a `class_eval`-of-a-string macro at its call sites

felixefelip:
## The gap

`has_rich_text :content` is three ordinary methods plus one `has_one`. The `has_one` is a reflection, so rbs_rails already types it:

```rbs
def rich_text_content: () -> ::ActionText::RichText?
def build_rich_text_content: (?untyped) -> ::ActionText::RichText
```

The three methods are written by `class_eval` on a heredoc **string**, and no static reader sees inside one — `ClassEvalExpander` desugars the *block* form, and felixefelip/steep#135 declines the string form for the same reason. So `post.content` is not `untyped` today, it is a **`NoMethodError`**: the method does not exist for the checker at all.

## What this emits

One file, per project rather than per model: `has_rich_text` as ActionText wrote it, sliced from the installed gem into `sig/generated/steep_actiontext_runtime/`. It states no type, and it does not generate the accessors — nothing does.

## The accessors are inferred, and the checker is what reads the macro

What the macro writes at a call site is a question about a **value**, and with the string-literal work of #345 landed this project answers those with the type machinery. A specialized body types `name` as `:content`, an interpolation over literals folds, and the folded literal **is** the source. `steep check` records it per call site in `sig/generated/.steep_string_evals.yml` (felixefelip/steep#169), and `RbsInfer::Project::StringEvalMacroExpander` places it in the class that made the call:

```ruby
class Article
  def content
    rich_text_content || build_rich_text_content
  end

  def content?
    rich_text_content.present?
  end

  def content=(body)
    self.content.body = body
  end
end
```

Measured against rbs_rails' output for the `has_one`, the model's own RBS gains:

```rbs
class Article < ApplicationRecord
  def content: () -> ActionText::RichText
  def content?: () -> bool
end
```

`content` because `rich_text_content || build_rich_text_content` is the union of the nilable reader and the non-nilable builder; `content?` because `.present?` is `bool`.

## The reader is gone

An earlier revision of this PR read the macro itself: `StringEvalMacro` walked the def for its `class_eval` strings and the conditions each sat under, `bind` matched a call site's arguments to the parameters, `guards_hold?` decided the branches and `render_parts` interpolated. A small interpreter over Ruby, sitting next to a checker that reads the same source semantically — the shape `docs/engineering/prefer-principled-over-shortcuts.md` names — and it cost the usual thing: a second reader has a second idea of "I cannot tell", and this one spelled it `false`, which is not a refusal but an answer voting for the `else` branch. `LiteralFold` (4353c16) fixed that and was reverted (1000f06) as the wrong shape, which #345 said it would replace.

So `string_eval_macro.rb`, `string_eval_macro_index.rb`, the binder, the guard walk and the interpolator are deleted (~400 lines). What is left decides only **where**, which is the one thing the sidecar cannot say. What that buys beyond the deletion:

- the branch is decided by narrowing, so `if`, `case`/`when` and a ternary all work, and an undecidable one cannot read as a decision;
- a call omitting an optional keyword takes the definition's default — how `has_rich_text :content` gets the `store_if_blank: true` writer while `has_rich_text :summary, store_if_blank: false` gets the other one;
- an interpolation is anything the checker folds: `#{name.to_s.upcase}`, a constant, a call that folds under felixefelip/steep#168;
- a chunk whose value the call site does not fix arrives as a hole and declines the whole call site, rather than being quietly absent.

That is what makes the Rails 8.1 `store_if_blank:` writer work without this generator knowing the option exists, and it is the core's doing, not the extension's: `class_eval` of an interpolated string is a plain-Ruby idiom, so an app that writes the same shape in its own concern gets the same treatment with no generator at all.

## `# @rbs_infer no-signature`

One new marker, and the transcription is its first user. gem_rbs_collection already declares `has_rich_text`, so the file cannot declare it plainly (`DuplicatedMethodDefinitionError` poisons the whole environment). An overloading `|...` is the wrong tool: it puts an inferred `(untyped name, …)` **ahead** of the gem's `(String | Symbol name, …)` — and a literal argument against an `untyped` parameter stops being a literal, which is the whole mechanism that reads the body. Declaring nothing leaves the accurate signature in place. The marker is core and general: pseudo-code transcribed to be *read*, whose signature is already written elsewhere.

## The mixin is the engine's own line

The transcription reproduces the shape ActionText writes — an `ActiveSupport::Concern` whose `class_methods do` block holds the macro — and mixes it in the way the engine does:

```ruby
class ActiveRecord::Base
  include ActionText::Attribute
end
```

The `extend ActionText::Attribute::ClassMethods` this PR used to state is gone: the Concern performs it, and the AR runtime's pseudo-code derives it (`base.extend const_get(:ClassMethods)`). The one liberty left is moving the `include` out of the `ActiveSupport.on_load(:active_record)` it is written in and onto the class that hook names — `OnLoadExpander`'s own table rather than a guess, and forced by #353.

Three defects were in the way, each reproduced before fixing:

- **The def was keyed by its lexical position.** `class_methods do` puts `has_rich_text` lexically in the concern while every call site resolves to its `ClassMethods`; the two names never met and the body was never specialized. Fixed in felixefelip/steep#169 by reading the block's `@implements`.
- **`# @rbs_infer no-signature` did not survive the relocation.** `BlockReopen.body_source` slices a block body from the first statement, dropping the comments above it — so *every* annotation on a relocated def (`#:`, `@rbs`, `@rbs_infer`) was silently lost. The walk up is over bytes now: Prism reports byte offsets and `String#rindex` counts characters, and one `—` in the generated header was enough to put it a byte off.
- **The derived `extend` doubled the written one.** `include X` implying `extend X::ClassMethods` is a derivation, and the Concern pseudo-code writes the same fact into the source; `RbsBuilder` emitted both. The written one wins.

What the shape buys, beyond being what the gem says: `ClassMethods` gets its `self` as `singleton(::ActiveRecord::Base) & ::ActionText::Attribute::ClassMethods`, derived from the concern's host the way every app concern's is, rather than stated by an `EXTENDER` constant the transcriber no longer needs.

## The cost## The cost

The expansion now needs one `steep check` ahead of it — the same cycle `.steep_postconditions.yml`, `.steep_contracts.yml` and `.steep_specializations.yml` already live in. A first run against a project with no sidecar renders no macro; the second renders it.

## Scope

The app-side accessors. `ActionText::RichText`'s own methods (`to_plain_text`, `to_trix_html`, the `delegate`s to `body`) are equally plain Ruby and equally transcribable, but they all read `body`, which rbs_rails types `::String?` — the column type — rather than `::ActionText::Content`. Its serializer handling special-cases only `ActiveRecord::Coders::JSON`, `Array` and `Hash` and falls through to `sql_type_to_class` for any other coder, so `serialize :body, coder: ActionText::Content` is typed as its column. Transcribing those bodies before that is fixed would emit code that reports an error instead of a type, so that half waits on the rbs_rails coder fix.

## Follow-up

The `include` is hoisted out of the `on_load` block rather than transcribed inside it, because `MixinIndex` reads raw sources and an `include` left in the hook reaches no host. That blind spot is #353 and is untouched here.

## Testing

- Unit: **1491 examples, 0 failures**. New specs for `StringEvalSidecar` (6) and a rewritten `StringEvalMacroExpander` spec (12) that exercises placement against a hand-built sidecar; reading the macro is Steep's half and is tested there (12 more).
- Integration: **157 examples, 0 failures**, including the `Article` snapshot, the expanded-source snapshot and the transcription snapshot. The rendered accessors are byte-identical to what the reverted reader produced — only the fixture's comments moved.
- `steep check` on the dummy holds its 53-diagnostic baseline (the four ActionText entries moved two lines and now name the intersection self type), and the dummy reaches a fixed point in two cycles.
- `Article` and `.steep_string_evals.yml` are byte-identical to what the stated `extend` produced — the shape changed, what it infers did not.
- Needs felixefelip/steep#169.

## Review round

Two findings, both reproduced before fixing and both in `830840b`:

- **An absolute constant path was folded into the lexical namespace.** `module Outer; class ::Article; has_slot :content; end; end` reopened `Outer::Article`, a class the program does not have. The `::` prefix now decides, for the reopen and for the nesting carried below it.
- **A call under control flow was dropped.** Only direct statements of a class body were consulted, so a macro call inside `if Rails.env.production?` defined nothing though it runs on the same `self`. The walk goes through `if`/`case`/`begin` and stops where `self` stops being the class: `def` bodies, nested classes and modules, blocks and lambdas.

Upstream, felixefelip/steep#169 gained the same round: a chunk is now claimed only where the call site decides it runs. The dummy does not move for either.

https://claude.ai/code/session_01UfeSTBcB41WMGTP9db8kNd


felixefelip:
## Review

The architecture is right: the core expander names no gem, the generator writes one transcribed file, and nothing states a type. The payoff (`content: () -> ActionText::RichText` with zero annotations) is real. But the conservatism the code advertises is not the conservatism it implements — below are four repros, run against this branch, where it emits methods that do not exist at runtime.

---

### 1. Blocker — "every uncertainty declines the macro" is not what the code does

`StringEvalMacro` (`lib/rbs_infer/project/string_eval_macro.rb:30-41`) promises that an undecidable condition declines. In practice **only `if`/`unless` with a bare parameter read** is understood. Everything else that gates a `class_eval` is treated as unconditional, or silently resolved to the `else` branch.

**a) A predicate that isn't a param read picks the `else` branch.**

```ruby
def slot(name)
  if fancy_mode?
    class_eval "def #{name}_fancy; end"
  else
    class_eval "def #{name}_plain; end"
  end
end
```

`slot :size` → emits `def size_plain`. It did not decline; it *decided*.

**b) Negation is invisible — wrong in both directions.**

```ruby
def slot(name, writable: true)
  if !writable then class_eval "def #{name}_ro; end"
  else              class_eval "def #{name}_rw; end" end
end
```

`slot :size, writable: false` → Ruby defines `size_ro`. The expander emits `size_rw`: the method that does not exist, and drops the one that does.

**c) `case`/`when` emits *every* branch.**

```ruby
case kind
when :ro then class_eval "def #{name}; end"
when :rw then class_eval "def #{name}=(v); end"
end
```

`slot :size` → emits both `size` and `size=`.

**d) `&&` short-circuit is ignored.**

`writable && class_eval("def #{name}=(v); end")` with `writable: false` → emits `size=`.

**Root cause, two spots:**

- `string_eval_macro_expander.rb:174-186` — `bound` returns `nil` both for *"this predicate is not a parameter read"* and for *"the parameter is bound to nothing"*, and `truthy?` maps `nil` to a definite `false`. That conflates "cannot read" with "is falsy", which makes the `:undecidable` machinery above it nearly unreachable and makes the `else` branch (polarity `false`) match every time.
- `string_eval_macro.rb:106-108` — the catch-all `else` recurses with **unchanged guards**, so `case`, `when`, `&&`/`||`, loops and `rescue` are all transparent.

Why the specs do not catch it: `spec/lib/rbs_infer/project/string_eval_macro_expander_spec.rb:157` ("declines a branch it cannot decide") tests an undecidable *value* (`writable: CONFIGURED` → `ConstantReadNode` → `:undecidable`). The undecidable *predicate shape* is never exercised.

Suggested shape of the fix:

```ruby
# expander
def guard_value(predicate, bindings)
  name = StringEvalMacro.parameter_read(predicate) or return :undecidable
  return :undecidable unless bindings.key?(name)
  truthy?(bindings[name])
end
```

and in `collect`, whitelist the transparent nodes (`StatementsNode`, `BeginNode`, `ElseNode`, the two conditionals) and push an `:undecidable` guard for anything else, so a `class_eval` reached through a `case`, an `&&` or a block is not known to run. That makes (a)–(d) all decline — which is what the docs already claim.

### 2. Soundness gap — a macro is matched by name alone, with no ancestor check

`expansions_for` (`string_eval_macro_expander.rb:82-90`) does `macros[stmt.name]` for any receiverless call in a class body. Nothing checks that the calling class actually has the defining module in its ancestors. The index's dedup rule (`string_eval_macro_index.rb:16-21`) argues the ancestor chain lives in RBS and therefore requires a *unique* name — but uniqueness only rules out picking the wrong macro; it does not establish that the caller has *any* macro. A single-definition name called by an unrelated class (a DSL from a gem that is not in the corpus, and so is not indexed) writes phantom methods onto it. Worth at least documenting as a known limit, since the rest of the file is explicit about its limits.

### 3. The vendored `.gem_rbs_collection` patch looks unnecessary

`spec/dummy/.gem_rbs_collection/actiontext/7.2/actiontext.rbs` is hand-edited to add `?store_if_blank: bool`. Reverting just that hunk and running `steep check app/models/article.rb` in the dummy gives **"No type error detected."** The `# @rbs_infer |...` overloading declaration the generator already emits accepts the keyword, so the stub edit is dead weight — and it is the kind of edit that silently disappears on an `rbs collection` refresh.

### 4. Smaller things

- **`StringEvalMacroIndex.none` (`string_eval_macro_index.rb:29-32`) is dead production code** — its only caller is its own spec (`string_eval_macro_index_spec.rb:79`). The comment justifies it at length via `docs/engineering/required-threaded-deps.md`, but nothing threads it. Delete it, or use it.
- **Unknown keywords at the call site are ignored** (`string_eval_macro_expander.rb:140-156`). `slot :size, writable: false` against `def slot(name)` renders happily; at runtime that is an `ArgumentError`. Same for a macro with no parameters at all — `bind` (`:120`) only rejects extra *positionals*.
- **`AttributeTranscriber` parses actiontext's `attribute.rb` three times per `file_entry`** — `macro_source`, then `engine_root` → `macro_def_node`, then `gem_file` → `macro_def_node` again (`attribute_transcriber.rb:127-146`). And `path = macro_def_node&.location && gem_file(...)` at `:128` is a redundant guard, since `gem_file` re-derives the same node. One memoized `macro_def_node` collapses all of it.
- **No cheap gate on the expander.** Once the ActionText sidecar exists, `macros.any?` is true for every project, so `expand` runs a full extra `Prism.parse` on **every** analyzed file (`analyzer.rb:101`). A one-line `macros.keys.any? { |n| source.include?(n.to_s) }` before the parse restores the "a project without the idiom pays nothing" property the index gets right.
- **Generator specs are coupled to the installed gem's text** (`expect(source).to include("if store_if_blank")`). The generator degrades gracefully on a Rails rewrite; the suite will not. Fine as a tripwire, but say so.

### 5. Diff hygiene

Roughly 1.5k of the 4.1k added lines are dummy-sig regeneration, and some of it predates this PR: `sig/rbs_infer/.expanded/lib/rails_ext/active_storage_authorization.rb` gains the four `included do` replays from **#342** (the parent commit), and `relation_prepend_order.rbs` is a snapshot that was never committed for that same file. Fine to absorb, but worth one line in the PR body so a reviewer does not hunt for what in *this* change produced them. Same for the four empty `class ActiveStorage::…Controller end` declarations that now duplicate the four reopens below them in `active_storage_authorization.rbs` — that is #342's output shape, not this PR's, but it becomes visible here.

Also: the PR body still describes the **previous** iteration ("What this emits: one reopen per model … `def content`"), which the head commit `04fad1c` replaced with "transcribe the macro, generate nothing". Worth rewriting before merge.

---

**Bottom line:** #1 is a correctness bug with four reproductions and a small fix; I would hold the merge on it. #2 and #3 are worth resolving in the same pass. The rest is cleanup.

https://claude.ai/code/session_01Myw5xJLtv8mqZoXLqTvjeW

felixefelip:
Finding #1 is fixed in 4353c16 — and not by patching `truthy?`, but by removing the reason there were two answers to begin with.

The two readers are now one function. `Project::LiteralFold` folds an expression to `String | Symbol | true | false | nil | UNKNOWN` against the call site's bindings, and both the interpolation (`render_parts`) and the condition (`guards_hold?`) go through it. `UNKNOWN` is its own object rather than a Symbol, because `slot :unknown` binds a parameter to `:unknown` and a sentinel a call site can spell is a sentinel that lies. There is no longer a place where "could not read" can be written as an answer.

The guard walk changed with it: **refusal is the default**. Only `StatementsNode` and `ElseNode` pass guards through unchanged; anything the walk does not recognise contributes `UNREADABLE`, which is neither true nor false. Chunks under a barrier are *collected*, not skipped — skipping one would drop it silently and let the macro's other chunks render a class with a reader and no writer. On top of that, `case`/`when` is now read as the chain of conditions it is, and `&&`/`||` as the condition the left side decides.

The four repros:

| shape | before | after |
|---|---|---|
| `if fancy_mode?` | emitted `size_plain` | declines |
| `if !writable`, `writable: false` | emitted `size_rw` | emits `size_ro` |
| `case`/`when` | emitted every branch | emits the selected branch |
| `x && class_eval(…)` | emitted unconditionally | emits only when `x` | 
| `class_eval` inside a block | emitted | declines |

Specs: `spec/lib/rbs_infer/project/literal_fold_spec.rb` (13 examples, including the `:unknown` collision) plus 7 in the expander spec — three under "reading the condition", four under "declining", one of which pins that a single unreadable chunk declines the *whole* macro.

Full suite green: **1693 examples, 0 failures**. No emitted output moves — the dummy's `Article` snapshot is unchanged, since Rails' own `if store_if_blank` was already inside the shape the old reader could decide.

Still open from the review above: the name-only macro match with no ancestor check (#2), the `.gem_rbs_collection` patch that `steep check` says is unnecessary (#3), and the smaller items in #4.

https://claude.ai/code/session_01Myw5xJLtv8mqZoXLqTvjeW

lib/rbs_infer/project/string_eval_macro_expander.rb
```diff
@@ -0,0 +1,134 @@
...
+        node.compact_child_nodes.each { |child| walk(child, namespace, path, sidecar, reopens) }
+        return
+      end
+
+      name = RbsInfer::Analyzer.extract_constant_path(node.constant_path)&.delete_prefix("::")
+      # A constant path this cannot name (`class self::Thing`) stops the walk:
+      # anything below it would be attributed to the wrong owner.
+      return unless name
+
+      qualified = (namespace + [name]).join("::")
```
felixefelip: **[P1] Preserve caminhos de constante absolutos**

Remover `::` antes de qualificar perde a informação de que a declaração ignora o namespace lexical. Reproduzi `module Outer; class ::Article; has_slot :content; end; end` gerando um reopen de `Outer::Article`. Detecte o prefixo absoluto e, nesse caso, não concatene `namespace`.

lib/rbs_infer/project/string_eval_macro_expander.rb
```diff
@@ -0,0 +1,134 @@
...
+
+        sources = sidecar.sources_for(
+          path: path,
+          line: stmt.location.start_line,
+          column: stmt.location.start_column
+        )
+        next unless sources
+
+        sources.map { |source| dedent(source) }.join("\n")
+      end
```
felixefelip, line 103: **[P2] Percorra controle de fluxo do corpo da classe**

Somente chamadas que são statements diretos são consultadas. Chamadas dentro de `if true`, `case` ou `begin` continuam executando com o mesmo `self` da classe e podem estar precisamente identificadas no sidecar, mas são descartadas aqui. Percorra nós estruturais executáveis mantendo o owner, parando em métodos e demais contextos onde `self` não possa ser determinado.

Issue #345, opened 2026-09-11
Literal types should survive a call site (Example69): specialize a method body per call site


felixefelip:
A literal a human can read off the source is dropped the moment it crosses a call site. `call(flag_name: false)` infers `String`, not `'delete'`, and the same loss is what keeps a `class_eval`-written macro unreadable — the motivating case, but not the mechanism.

**Design note (private artifact, full evidence and numbers):** https://claude.ai/code/artifact/50507797-3d38-4c2e-a9d4-3f0bd9026694

## The fixture is the spec

`Example69` (`spec/dummy/app/models/example69.rb`) is plain Ruby with no macro in it, and states beside each method the literal it should infer. It is wired into the suite on `feat/literal-types` with today's baseline — every `String` in `spec/expectations/models/example69.rbs` is a place a literal is known and dropped.

Five distinct failure modes, which is what earns it a snapshot:

| expected line | catches an implementation that |
|---|---|
| `call_name_action: () -> 'name_delete'` | specializes nothing |
| `call_dynamic: () -> 'name_delete'` | specializes a single hop |
| `name_action_dynamic: (flag_name: bool) -> ('name_delete' \| 'delete')` | over-specializes the DECLARATION from one call site |
| `call_unknown: (flag_name: untyped) -> ('name_delete' \| 'delete')` | picks a branch it cannot decide |
| `call_both: () -> Array['name_delete' \| 'delete']` | loses the literal into a collection |

`name_action_dynamic` and `call_unknown` have the same body and differ only in having callers — the fixture's own premise, which rbs_infer already states today (`bool` against `untyped`) without any of the machinery being asked for.

## What blocks it, in order

1. **The argument loses its literal.** `type_construction.rb`, the `:true`/`:false` case: with a hint that accepts them they are typed `unwrap(hint)` — the *hint*, not the literal. With `?flag_name: bool` declared, the `false` at the call site is already `bool` before anything else runs. The branch immediately below keeps the literal in condition position, so the checker knows how; it declines to when a hint is present.
2. **The interpolation dropped it.** `:dstr` synthesized its children for side effects and typed itself `::String`. Fixed in felixefelip/steep#163 — folds when every part carries a literal type, `::String` otherwise. 499 runs, 0 failures.
3. **The body is checked once.** `for_new_method` takes parameter types from the RBS declaration and `check`s the body against the declared return. There is no notion of "this body, with *these* argument types".

## Both primitives already exist

[...]

Issue #191, opened 2026-08-07
A generated return type is a ratchet: only `-> untyped` is ever revisited


felixefelip:
`ReturnTypeResolver#improve_method_return_types` asks Steep only about methods whose signature is still `-> untyped` (`still_untyped`). Once a concrete return type is written it survives every regeneration — including after the type it was derived from changes underneath it, which leaves a `Ruby::MethodBodyTypeMismatch` that regenerating cannot clear.

## Where it surfaced

Fizzy's `app/controllers/concerns/filter_scoped.rb`:

```
Cannot allow method body have type `((::Filter & ::Filter::Validated) | ::Array[::Filter])`
  because declared as type `(::Filter & ::Filter::Validated)`
```

The chain:

1. `ActiveRecord::Relation#build` gained an `(Array[untyped]) -> Array[Model]` overload in the collection. With an `untyped` argument every overload matches, so the result is the union.
2. `Filter.from_params` (`find_by_params(params) || build(params)`) went from `-> untyped` to `-> ((Filter & Filter::Validated) | Array[Filter])`.
3. `FilterScoped#set_filter` assigns from `from_params` in one branch, so its body now types as that union — but its RBS still declares the pre-widening `-> (Filter & Filter::Validated)`.

The regeneration that widened the ivar (`@filter`, recomputed from scratch every run) left the return alone, in the same file, in the same pass. Asked directly, Steep gets it right:

```
method_return_types → {"set_filter" => "((Filter & Filter::Validated) | Array[Filter])"}
ivar_write_types    → {"filter"     => "((Filter & Filter::Validated) | Array[Filter])?"}
```

Reproduced with no Rails involved — a 3-file project, one `if/else` assigning from a narrow and a wide method:

| seeded `sig/generated/picker.rbs` | analyzer output |
|---|---|
| absent | `-> ((Filter & Filter::Validated) | Array[Filter])` ✔ |
| `-> untyped` | `-> ((Filter & Filter::Validated) | Array[Filter])` ✔ |
| `-> (Filter & Filter::Validated)` | unchanged ✘ |

## Cause

Three loops after the `still_untyped` one do revisit an already-typed method, but each refines a declaration that is merely *vague* and that the body still satisfies: a block's `Array[untyped]`, a nilable Steep proves non-nil, a record with `untyped` values. None of them revisits a declaration the body **contradicts**, and the previously generated RBS is read back as a known return type, so the wrong answer is self-perpetuating.

## Fix

[...]

PR #349, opened 2026-09-12
feat(literals): take the body's literal over a declaration that only accepts it (S4 of #345)

felixefelip:
Consumer half of stage S4 of #345. Pairs with felixefelip/steep#166, which records the per-argument-tuple return types — merge that first.

## The symptom

With the checker side in place, `SteepBridge` loads `sig/generated/.steep_specializations.yml` and `method_return_types` already answers what a human reads off `Example69`:

```
call_name_action => "name_delete"
call_action      => "delete"
call             => ("name_delete" | "delete")
```

The emitted RBS said `String` for all three.

## The one place that decided it

`improve_method_return_types` applies the type `known_return_types` resolves from the **declaration** — the one the previous run wrote — and defers to the body only when `body_contradicts?`, which asks `accepts?`. A `String` accepts `"name_delete"`, so the declaration won, the member stopped being `untyped`, and the Steep pass holding the literal was never asked. Trace: `[apply] call <- "String"` while the bridge was saying the union.

That made the loss permanent, not just per-run: the declaration the literal lost to is the previous run's own output, so no number of `--max-passes` reaches it. Same ratchet #191 described, on a different axis.

## The rule

A fifth slice in the sequence that file already runs (a block's `Array[untyped]`, a nilable Steep proves non-nil, a record with `untyped` values, a body that says `self`, a contradicted declaration). This one: the declaration is right and the body satisfies it, but the body fixes the **value**.

`SteepBridge#literal_refinement?` holds it to that axis — the two types must be the same *up to literals*, with the body strictly more precise. Every condition is something the dummy did without it:

| condition | what it prevents |
|---|---|
| literals widened on **both** sides | by pass 2 the declaration is itself literal; `"name_delete"` vs `("name_delete" \| "delete")` must compare, or the first specialized call site is also the last |
| equivalence, not subtyping | `value <: declared` also holds for a narrower class (`Integer` vs `Numeric`) and a dropped nil — slices that already exist |
| strictly more precise | `("a" \| "b")?` was being rewritten as `("a" \| "b" \| nil)`: churn across a whole `sig/` for a type that did not change |
| `true`/`false` excluded | `bool` denotes exactly `(true \| false)`, so the literals buy no precision and cost two things: `Assignment#assigned?` came out `-> (true \| false)` and **lost the postcondition that proved `post`**, and unions grew noise (`(User \| bool \| true)`) |

`literal_bearing?` and `widen_literals` are written over `each_child`/`map_type`, which every Steep type answers (a leaf through `Helper::NoChild`), so the only type named is the literal itself. `accepts?` now shares the new `subtype?` helper.

## Result

`Example69` closes **8 of its 10 lines**:

```rbs
def name: () -> "name"
def action: () -> "delete"
def call_name_action: () -> "name_delete"
def call_action: () -> "delete"
def call_both: () -> Array[("name_delete" | "delete")]
def name_action_dynamic: (flag_name: bool) -> ("name_delete" | "delete")
def call_unknown: (flag_name: untyped) -> ("name_delete" | "delete")
def call: (?flag_name: bool) -> ("name_delete" | "delete")
```

`call_dynamic` and `call_dynamic_with_name` stay at the union where the fixture asks for a single literal. They need the second hop — stage S5 — and that is measured, not assumed: re-running `steep check` with the improved declarations grows no new tuple in the sidecar, because the runner's pass 2 reads declarations rather than pass 1's specializations.

Elsewhere in the dummy, 20 files gain a literal the program fixes — `"John Doe"?`, `42?`, `"all"`, `"bazingado"`, `{ body: "post_notification", ... }`, and a generated store accessor (`channel: () -> "none"?`). `bool` is untouched everywhere and `.steep_postconditions.yml` has no diff.

## Verification

- suite: **1646 examples, 0 failures** (7 new in `steep_bridge_spec.rb`, each negative case being one of the guards above)
- dummy reaches a fixed point in **two** `rbs_infer` → `steep check` cycles
- `steep check` holds its **49-diagnostic baseline**; the refreshed baseline is 7-for-7, the same messages now naming a literal type (`("John Doe" | nil)` where it read `(::String | nil)`)

https://claude.ai/code/session_01UfeSTBcB41WMGTP9db8kNd

### DIFF
--- a/lib/rbs_infer/inference/return_type_resolver.rb
+++ b/lib/rbs_infer/inference/return_type_resolver.rb
@@ -230,24 +230,11 @@ module RbsInfer::Inference
             m.signature = m.signature.sub(/-> #{Regexp.escape(current_type)}$/, "-> self")
           end
 
-          # A fifth slice, on the axis none of the others can reach: the
-          # declaration is right and the body satisfies it, but the body fixes
-          # the VALUE — `call(flag_name: true)` returns `"name_delete"` where
-          # the declaration of `call` says `String` (felixefelip/rbs_infer#345,
-          # stage S4, where Steep specializes a body per argument tuple).
-          #
-          # Nothing else writes it: the first loop only fills `-> untyped`, and
-          # the general case below needs the declaration to REJECT the body,
-          # which a `String` never does for its own literals. So the literal
-          # would be lost on every run — and permanently, because the
-          # declaration it loses to is the one the previous run emitted.
-          #
-          # `literal_refinement?` holds it to that axis: the body type must be
-          # the declared type with literals in place of their classes, so a
-          # narrower class or a dropped nil still belongs to the loop that owns
-          # it. `defines_own_body?` for the reason the first loop states — a
-          # second body under this name makes Steep's name-keyed answer
-          # ambiguous.
+          # A fifth slice: the declaration is right and the body satisfies it,
+          # but the body fixes the VALUE — `call(flag_name: true)` returns
+          # `"name_delete"` where `call` is declared `-> String`
+          # (felixefelip/rbs_infer#345). The general case below cannot reach it,
+          # because a `String` never rejects its own literals.
           members.each do |m|
             next unless method_member?(m)
             next if m.name == "initialize"
--- a/lib/rbs_infer/inference/return_type_resolver.rb
+++ b/lib/rbs_infer/inference/return_type_resolver.rb
@@ -230,6 +230,48 @@ module RbsInfer::Inference
             m.signature = m.signature.sub(/-> #{Regexp.escape(current_type)}$/, "-> self")
           end
 
+          # A fifth slice, on the axis none of the others can reach: the
+          # declaration is right and the body satisfies it, but the body fixes
+          # the VALUE — `call(flag_name: true)` returns `"name_delete"` where
+          # the declaration of `call` says `String` (felixefelip/rbs_infer#345,
+          # stage S4, where Steep specializes a body per argument tuple).
+          #
+          # Nothing else writes it: the first loop only fills `-> untyped`, and
+          # the general case below needs the declaration to REJECT the body,
+          # which a `String` never does for its own literals. So the literal
+          # would be lost on every run — and permanently, because the
+          # declaration it loses to is the one the previous run emitted.
+          #
+          # `literal_refinement?` holds it to that axis: the body type must be
+          # the declared type with literals in place of their classes, so a
+          # narrower class or a dropped nil still belongs to the loop that owns
+          # it. `defines_own_body?` for the reason the first loop states — a
+          # second body under this name makes Steep's name-keyed answer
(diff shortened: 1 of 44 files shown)


lib/rbs_infer/inference/return_type_resolver.rb
```diff
@@ -230,6 +230,48 @@ def improve_method_return_types(members, attr_types, parsed_target: nil)
             m.signature = m.signature.sub(/-> #{Regexp.escape(current_type)}$/, "-> self")
           end
 
+          # A fifth slice, on the axis none of the others can reach: the
```
felixefelip: Remova esse e todos os comentários adicionados que não forem úteis, siga as normas do Clean Code sobre comentários

lib/rbs_infer/signatures/steep_bridge.rb
```diff
@@ -117,11 +117,51 @@ def accepts?(declared, value)
       value_type = context_free_type(value, subtyping)
       return nil unless declared_type && value_type
 
-      subtyping.check(
-        Steep::Subtyping::Relation.new(sub_type: value_type, super_type: declared_type),
-        self_type: nil, instance_type: nil, class_type: nil,
-        constraints: Steep::Subtyping::Constraints.empty
-      ).success?
+      subtype?(value_type, declared_type, subtyping)
+    rescue StandardError
```
felixefelip: Má prática de tratamento de erros

lib/rbs_infer/inference/return_type_resolver.rb
felixefelip: Done in 53788fe — that block went from 18 lines to 4, and I cut the added comments across the bridge and the spec too (the `subtype?` and `widen_literals` ones entirely, since the code says it). What stayed is only what the code cannot: the issue reference, the `bool`/`(true | false)` exclusion, and why the unknown-name check is asked up front. The reasoning that was in prose is in the PR body and the commit messages.

lib/rbs_infer/signatures/steep_bridge.rb
felixefelip: Fair — it was an exception used as control flow, and it hid everything else with it. Fixed in 53788fe.

I checked what actually raised: only an alias the environment does not define, from `RBS::DefinitionBuilder#expand_alias2` ("Unknown alias name: ::not"). Wrong arity, an unknown class name and a bad record all return false without raising. So `context_free_type` now refuses a type whose names the env does not define (`RBS::Environment#type_name?`) — the same answer it already gave for an unparsable or context-dependent type — and both predicates dropped the `rescue` entirely, so an unexpected failure surfaces instead of becoming a silent nil.

One behaviour change worth flagging: a declaration naming a type the env does not know now answers nil ("cannot decide") where it answered false. Callers only correct on false, so that is the conservative direction, and the dummy is byte-identical. Suite green (1645), dummy still a fixed point, steep check still at 49.

PR #352, opened 2026-09-13
feat(literals): infer pure core call results

### LINKED ISSUE #351: Fold pure core calls over literal values into literal return types
## Problem

Literal types now survive project-defined call sites through the S4/S5 specialization work in #345, but they are still widened when the transformation happens inside a core method whose body is unavailable to the project analysis.

`Example49` already contains both minimal reproductions:

```ruby
def self.call_color
  color.upcase       # color: "yellow"
end

def self.call_age
  age + 10           # age: 31
end
```

Current output:

```rbs
def self.call_color: () -> String
def self.call_age: () -> Integer
```

Expected output:

```rbs
def self.call_color: () -> "YELLOW"
def self.call_age: () -> 41
```

The literal receiver is widened to its backing class for method lookup, and the selected core RBS declaration can only answer `String#upcase -> String` or `Integer#+ -> Integer`. S4/S5 cannot recover the value because the core implementation has no Ruby body in the checked project.

## Proposed direction

Add a closed `LiteralIntrinsics` evaluator in Steep at the typed-send boundary, after argument checking and overload selection—the same point where S4 currently replaces a call's return with a specialization.

Conceptually:

```text
validated call + selected overload
              |
apply S4 specialization, if any
              |
LiteralIntrinsics.fold(call, receiver_type, argument_types)
              |
verify folded type <: declared return type
              |
call.with_return_type(literal)
```

This must be semantic, keyed by the fully resolved method identity, rather than syntax such as `method_name == :upcase`. If folding declines for any reason, preserve the selected RBS return unchanged.

No new inference logic should be needed in `rbs_infer`: its existing `literal_refinement?` pass already replaces a wider emitted return with the literal Steep reports.

## Initial closed set

Only exact, supported arities with a single literal receiver and literal argument types:

- `String`: `upcase`, `downcase`, `capitalize`, `reverse`, zero-argument `strip`, `+` with a literal String, `*` with a literal Integer, `length`, `to_sym`.
- `Integer`: `+`, `-`, `*`, `/`, `%`, `**` with a literal Integer, plus `abs`, `succ`, and initially zero-argument `to_s`.
- `Symbol`: `to_s`; case transformations can follow separately.

Do not include bang methods, I/O, blocks, splats, unknown arguments, project methods, or results RBS cannot represent as literals.

Methods already declared `-> self`, such as `String#freeze` and `Symbol#to_sym`, should not need this evaluator. In particular, `freeze` must not be classified as pure because it mutates object state even though its return value aliases the receiver.

## Core provenance and overrides

Checking only `call.method_decls` is insufficient to rule out monkey patches. It proves the RBS call resolved to a method named `::String#upcase`, but a Ruby source reopen:

```ruby
class String
  def upcase
    "something else"
  end
end
```

has the same owner and name. `method_decls` describes the RBS interface, not the runtime body selected from the project's Ruby sources.

Folding therefore also needs a project-wide source-defined-method/override registry, or an equivalent conservative guard. If the checked corpus can replace a candidate intrinsic through a direct definition, `prepend`, a modeled include, or a refinement, decline folding for that method. A first implementation may conservatively disable intrinsics for a core class whenever its lookup has been altered in a way the registry cannot resolve precisely.

The evaluator must never dispatch an arbitrary method name with `public_send` or `eval`. If it invokes the checker's Ruby core implementation, use closed handlers and verify that the captured implementation is genuinely native; otherwise decline. This matters because application sources are parsed rather than executed, while code loaded by the checker process can still patch core classes.

## Bounded and protected evaluation

Rescuing exceptions and widening an oversized result afterward is not enough:

```ruby
"x" * 1_000_000_000
10 ** 1_000_000_000
```

can exhaust memory or CPU before a result exists.

Each allocating or super-linear intrinsic needs a preflight cost/size bound. Apply a second bound to the serialized result and widen it to its backing class when it exceeds the literal budget. The budget can share the S5 widening vocabulary, but it must cover both computation cost and output representation.

Expected runtime failures such as division/modulo by zero, negative string repetition, invalid bases, incompatible encodings, or unsupported result classes should make folding decline without hiding the ordinary type-check result.

Unicode case conversion follows the checker Ruby version. Keep the initial set limited to operations independent of system locale and document that the result is relative to the Ruby core semantics used by the checker.

## Acceptance criteria

- `Example49::Bar.call_color` infers `"YELLOW"`.
- `Example49::Bar.call_age` infers `41`.
- A literal result continues through a larger chain or collection.
- Dynamic receivers or arguments retain the nominal RBS return.
- A project redefinition of a candidate core method disables folding.
- Bang methods are never folded.
- Division by zero and other exceptional inputs do not crash inference.
- Oversized string/numeric results are rejected or widened before expensive allocation.
- The folded literal is accepted only when it is a subtype of the selected declaration's return.
- Steep has focused tests for every supported family and every refusal path.
- The `rbs_infer` dummy snapshot covers the end-to-end Example49 result.

## Repository split

The evaluator and its safety/provenance machinery belong in the Steep fork, where the typed send and selected declaration are available. This repository should only need the integration fixture/snapshot update once the Steep change lands.

Related: #345 (project-defined literal specialization and fixpoint widening).

felixefelip:
## Summary

- feed the complete project source corpus into Steep core-literal folding
- keep each bridge registry isolated so one expanded source cannot taint later targets
- infer exact returns such as `"hello".upcase -> "HELLO"` and `31 + 10 -> 41`
- update generated signatures, sidecars, snapshots, and the Steep diagnostic baseline to the literal fixpoint

`Example49::Bar.call_color` now returns `"YELLOW"` and `call_age` returns `41`.

## Dependency

Depends on felixefelip/steep#168.

## Tests

- `make test_parallel` — 1,641 examples, 0 failures
- Rails dummy snapshots — 142 examples, 0 failures
- Steep dummy baseline — 2 examples, 0 failures
- focused cross-file override and bridge isolation specs

Closes #351


felixefelip:
Atualizei o PR com o bloqueio de dependência apontado na revisão do Steep: `SteepBridge#initialize` agora exige `literal_method_registry:` e todos os callers/specs independentes passam o registry explicitamente (`50ef0a2`).

Validação local dos consumidores afetados: 136 exemplos, 0 falhas.

